Issue #23, opened 2015-04-16
Support for sparse matrices


scheidan:
Thanks for providing this extremely useful package!

I was wondering if it would be reasonable to include sparse covariance matrices. I'd be happy to help, although my knowledge about sparse algorithm is very limited.

lindahua:
I think sparse covariance is useful. If you are interested in implementing this, PR is welcomed!

PR #24, opened 2015-04-24
Sparse covariance matrices

scheidan:
A type and the required functions for sparse covariance matrices as mention in #23. However, I need some help with two points:
-  It passes the `test_pdmat` test, however, not the addition tests.  Should the addition methods be derived automatically or do we have to add the new methods in `addition.jl` manually?

- I'm not sure if all calculations in `pdsparsemat.jl` are done in the most efficient way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,11 @@ os:
   - osx
 julia:
   - release
-  - nightly 
+  - nightly
 notifications:
   email: false
-# uncomment the following lines to override the default test script
-#script:
-#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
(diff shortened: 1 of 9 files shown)


lindahua:
In terms of add, you may at least add a method that implements addition between this type of matrix and ``PDMat``.

scheidan:
Added the addition between the other `PDMats` types. 

Maybe we should update the `README.md` as it currently implies that the addition methods are derived automatically for a new `AbstractPDMat` subtypes.

I'll investigate why it fails with the nightly build.

scheidan:
works now with Julia 0.3 and the latest nightly

lindahua:
Thanks! Looks great.

PR #32, opened 2015-09-30
Added __precompile__ flag to help with using on small-memory VMs.

janmtl:
In trying to use Distributions.jl on an HPC cluster, I found that I need to precompile each of my dependencies one-by-one. Adding this flag allowed me to force a precompile on PDMats.jl so that the subsequent precompiles actually ran inside the memory limit.

### DIFF
--- a/src/PDMats.jl
+++ b/src/PDMats.jl
@@ -1,4 +1,4 @@
-__precompile__(true)
+VERSION >= v"0.4.0-dev+6521" && __precompile__(true)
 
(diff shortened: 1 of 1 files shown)


janmtl:
Needed to add a version check before the __precompile__ flag. I used to one from Distributions.jl.

andreasnoack:
@janmtl Thanks for adding this. The error is a general error we have on Travis with Julia nightlies.

PR #34, opened 2015-12-24
Fix chol depwarn

yuyichao:
Ref JuliaLang/julia#13680

@andreasnoack


andreasnoack:
Do you know why the lower version is needed in this package? Usually, it shouldn't really matter.

yuyichao:
Not really

PR #37, opened 2016-01-27
Supportfloat32

KrisDM:
I've made PDMats typed on T<:AbstractFloat and removed all hard-coded references to Float64. 

It's all working cleanly for PDMat, PDDiagMat and ScalMat. A few workarounds of existing problems were needed to get it to work with PDSparseMat.

Reason why I did this is so that in the Distributions.jl package it will be easier to allow distributions to generate different floating types (for instance, the multi-variate normal function uses PDMat internally).

Can someone review and let me know if this is good to be merged into the main branch?

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,15 +18,21 @@ Positive definite matrices are widely used in machine learning and probabilistic
 
 ## Positive definite matrix types
 
-This package defines an abstract type ``AbstractPDMat`` as the base type for positive definite matrices with different internal representations.
+This package defines an abstract type ``AbstractPDMat{T<:AbstractFloat}`` as the base type for positive definite matrices with different internal representations.
+
+Elemenent types are in princple all AbstractFloat types, but in practice this is limited by the support for floating point types in Base.LinAlg.Cholesky.
+  - ``Float64``     Fully supported from Julia 0.3.
+  - ``Float32``     Fully supported from Julia 0.4.2. Full, diagonal and scale matrix types are supported in Julia 0.3 or higher.
+  - ``Float16``     Promoted to ``Float32`` for full, diagonal and scale matrix. Currently unsupported for sparse matrix.
+  - ``BigFloat``    Supported in Julia 0.4 for full, diagonal and scale matrix. Currently unsupported for sparse matrix.
 
 * ``PDMat``: full covariance matrix, defined as
 
 ```julia
-immutable PDMat <: AbstractPDMat
+immutable PDMat{T<:AbstractFloat,S<:AbstractMatrix} <: AbstractPDMat{T}
     dim::Int                    # matrix dimension
-    mat::Matrix{Float64}        # input matrix
-    chol::Cholesky{Float64}     # Cholesky factorization of mat
+    mat::S                      # input matrix
+    chol::Cholesky{T,S}         # Cholesky factorization of mat
 end
 
 # Constructors
@@ -37,9 +43,6 @@ PDMat(mat)          # with the input matrix, of type Matrix or Symmetric
                     # Remarks: the Cholesky factorization will be computed
                     # upon construction.
 
-PDMat(mat, uplo)    # with the input matrix, and an uplo argument (:U or :L)
-                    # to specify the way Choleksy is done
-
 PDMat(chol)         # with the Cholesky factorization
                     # Remarks: the full matrix will be computed upon
                     # construction.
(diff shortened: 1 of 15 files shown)


KrisDM:
Summary: 
-in v0.4.2 and above, all types support both Float64 and Float32
-in v0.3 up to 0.4.1, all types except one support both Float64 and Float32. The remaining type, PDSparseMat only supports Float64 because of the cholesky factorization functions it relies on.

KrisDM:
I noticed there would still be some incompatibilities with typealiases in the Distributions package, so Distributions would have to pin to 0.3.6 of PDMats until the PR I am working on there has been implemented.

kmsquire:
The proper way to handle this would be to set a maximum PDMats version in Yhe Distributions.jl requires files (in Metadata.jl) when PDMats is tagged, for all previous versions of Distributions. 

Cc:@tkelman

tkelman:
If something works in 0.4.2 but not in 0.4.1 or 0.4.0, then the lower bound should also be specific about the patch version. Otherwise people on old 0.4 releases get broken code when they do Pkg.update.

KrisDM:
I may have trouble finding out the specific patch version. It's not the code in this package itself that's broken in 0.4.1, it would run perfectly fine with Float64, so everything that worked before this PR would still work after.

For the newly supported type (Float32) in 0.4.1, a call to Base.LinAlg.cholfact generated a stack overflow in essentials.jl, which was fixed at some point but I don't know as a result of which issue. 

I was developing in 0.4.1 when I ran into the stack overflow, updated to 0.4.3 and it was gone. So didn't open an issue and also have little idea what caused the problem.

How would I find out?

tkelman:
You can test against specific patch versions on travis (cleanest to do so on a different branch in your fork). Are you adding tests that would fail in older julia versions? I guess if no previously functioning code gets broken then it's safer, but seeing Pkg.test start failing would likely be surprising.

KrisDM:
Testing a specific patch version in travis may help, but finding the right patch number to test would still be a needle in a haystack, wouldn't it?

I made sure that Pkg.test passes for 0.3 on Travis by skipping the tests of PDSparseMat and Float32 in versions before 0.4.2. As the root cause of the error for PDSparseMat is somewhere in Base, there would be no way for me to get them to pass. But I do run the tests for other PDMat types and Float32 and they all pass on 0.3 on Travis. 

So given that this PR doesn't break any existing functionality, but adds new functionality not all of which will work in 0.3 or 0.4.0 or 0.4.1, is not having the correct patch number still a problem?

tkelman:
Sorry should have looked at the code, if you have the tests gated on version so they still pass on all 0.3.x and all 0.4.x, I think you can leave the lower bound alone. The upper bound issue @kmsquire mentioned is a separate problem and would still need to be handled.

KrisDM:
OK, thanks.

For the upper bound, how should I proceed? Does that need a separate PR for Metadata.jl, which needs to be approved before this one?

This is my first contribution to Julia packages, so don't yet know the working practices. Is it documented somewhere?

tkelman:
There's a section in the package docs on the require version specification. Technically anyone can make that metadata pr. It would involve adding an upper bound of 0.4 (upper bounds are exclusive) to all past versions of Distributions' requires files on the PDMats dependency line. Maybe other packages that depend on this one too if you think it might break anything else? Good to check first, a `git grep PDMats` in METADATA can tell you which packages depend on this one.

Then the first tag of PDMats with this change would be 0.4.0 of the package, and your distributions fix could be tested against this version.

KrisDM:
Thanks, will read docs and check other dependencies.

KrisDM:
@kmsquire @tkelman I have created the following PRs to fix the upper bound issue for packages relying on PDMats. After doing ``git grep PDMats`` and testing all packages that showed a dependency, only Distirbutions.jl and ConjugatePriors.jl needed their upper bound fixed (for the other packages, the test either passes, or the packages are not in development and broken for other reasons).

Here are all corresponding PRs:
(1) In METADATA,jl: https://github.com/JuliaLang/METADATA.jl/pull/4506
(2) In Distribuions.jl: https://github.com/JuliaStats/Distributions.jl/pull/458
(3) In ConjugatePriors.jl: https://github.com/JuliaStats/ConjugatePriors.jl/pull/9

Does either of you have write access to approve these and then approve this PR?

Thanks!

tkelman:
Wonderful, looks great. I can merge the METADATA PR, but don't have commit access in JuliaStats.

kmsquire:
I just merged the commits to Distributions.jl and ConjugatePriors.jl... and realized that I jumped the gun a little and should have waited until this one was ready.  Whoops!

It looks good, but it would be great if you could squash the commits.  As soon as you do that, I'll merge.  Ask back here if you're uncertain how to do that.

kmsquire:
(I can always revert those other commits, if needed.)

kmsquire:
Okay, never mind, it was obviously fine to commit those. I also need to pay more attention.  @KrisDM, thanks for jumping through these hoops!

KrisDM:
@kmsquire I did the squashing/rebasing. In this case it turned out to be simple enough for a git newbie:-). 

So I suppose that, once merged, this will become available as PDMats v"0.4.0"? 

Once it's in, I'll start preparing the fixes for the Distributions package so that I can start using the newer functionality over there.

README.md
```diff
@@ -49,41 +52,43 @@ PDMat(chol)         # with the Cholesky factorization
 * ``PDiagMat``: diagonal matrix, defined as
 
 ```julia
-immutable PDiagMat <: AbstractPDMat
+immutable {T<:AbstractFloat,V<:AbstractVector} <: AbstractPDMat{T}
```
kmsquire: Missed this earlier--missing `PDiagMat`

kmsquire: (You don't have to squash again, since this is just a doc fix.)

src/scalmat.jl
```diff
@@ -1,14 +1,12 @@
...
-    dim::Int
-    value::Float64
-    inv_value::Float64
-
-    ScalMat(d::Int, v::Float64) = new(d, v, 1.0 / v)
-    ScalMat(d::Int, v::Float64, inv_v::Float64) = new(d, v, inv_v)
+immutable ScalMat{T<:AbstractFloat} <: AbstractPDMat{T}
+  dim::Int
+  value::T
+  inv_value::T
```
kmsquire: Should use 4 spaces.

src/pdsparsemat.jl
```diff
@@ -1,20 +1,20 @@
 # Sparse positive definite matrix together with a Cholesky factorization object
-immutable PDSparseMat <: AbstractPDMat
-    dim::Int
-    mat::SparseMatrixCSC{Float64}
-    chol::CholTypeSparse
+immutable PDSparseMat{T<:AbstractFloat,S<:AbstractSparseMatrix} <: AbstractPDMat{T}
+  dim::Int
+  mat::S
+  chol::CholTypeSparse
+  PDSparseMat(d::Int,m::AbstractSparseMatrix{T},c::CholTypeSparse) = new(d,m,c) #add {T} to CholTypeSparse argument once #14076 is implemented
```
kmsquire: 4 spaces.

kmsquire:
The requested changes don't change any functionality, so they can be addressed in a future PR if you'd prefer. I'll merge this tonight either way. Nice job!

KrisDM:
Thanks for checking, Kevin. 

I'll do those changes in next commit. Sorry about the 4 spaces, I think the settings in my Juno are set to 2, will change that so it is doesn't mess up formatting again.

Cheers!

tkelman:
Juno really needs to change its default indent to fit with the standards used in base and most packages.

PR #38, opened 2016-02-09
Rewrite colwise_dot! and colwise_sumsq! to use explict loops

simonster:
For small to moderately sized covariance matrices, the time cost of allocating these views can be pretty big. This was also the only place this package used ArrayViews.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,2 @@
 julia 0.3
-ArrayViews 0.4.8-
 Compat 0.7.7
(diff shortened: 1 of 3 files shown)


kmsquire:
LGTM.

kmsquire:
Actually, there aren't any tests for these functions--can you add something small?

simonster:
It's called by `quad!` and `invquad!`, which should have tests that are testing it, although it seems Coveralls was not enabled on this repository. (I just enabled it.) I'm not sure it really needs tests of its own; those tests should be sufficient and it's not exported or intended for external use anyway.

kmsquire:
Okay by me to merge, then.

simonster:
Interestingly, according to Coveralls, `utils.jl` has 109.8% code coverage.

PR #41, opened 2016-03-23
generalize pdmats to work with ForwardDiff

jmxpearson:
Relaxing the parameter type to `T <: Real` would allow PDMats (and thus eventually distributions using PDMats) to make use of automatic differentiation (à la ForwardDiff.jl).

### DIFF
--- a/src/PDMats.jl
+++ b/src/PDMats.jl
@@ -43,7 +43,7 @@ module PDMats
 
     # The abstract base type
 
-    abstract AbstractPDMat{T<:AbstractFloat}
+    abstract AbstractPDMat{T<:Real}
 
     # source files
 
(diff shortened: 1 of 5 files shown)


andreasnoack:
This is the right direction. I took a brief look and I think that most of the element type restrictions can be completely removed from the methods. It's fine that the type has a restriction but the methods should only be restricted when it's necessary for correct dispatch or when the body exploits specific features of the element type.

Would you mind trying to delete as many of the restructions as possible in this PR?

jmxpearson:
I think I got all the ones where the type parameter wasn't needed for the function body.

andreasnoack:
Thanks. Last thing: please squash your commits into a single. Then we can merge.

jmxpearson:
Done!

PR #42, opened 2016-03-28
generalizing scalmat

jmxpearson:
- allows scalmat to work with automatic differentiations
- missed in #41


andreasnoack:
What about just deleting this type? Isn't `UniformScaling` in base exactly the same?

jmxpearson:
Probably possible to remove it from PDMats, but `ScalMat` has a `dim` field, so there might be some breakage of downstream code.

andreasnoack:
Let's keep it. You are probably right. `UniformScaling` cannot be used as a covariance matrix for this reason.

PR #45, opened 2016-06-01
Loosen type restrictions

### LINKED ISSUE #44: Looser typing for utility functions
Ran into this while looking at autodiff for the `MvNormal` pdf, which calls `sqmahal`, which calls `quad`, which doesn't accept an array and vector with different eltypes.

I would propose replacing type signatures in `utils.jl` like

```julia
_addscal!{T<:Real}(r::Matrix{T}, a::Matrix{T}, b::Union{Matrix{T}, SparseMatrixCSC{T}}, c::T)
```

with 

```julia
_addscal!(r::Matrix, a::Matrix, b::Union{Matrix, SparseMatrixCSC}, c::Real)
```

and making sure non-inplace versions in `generics.jl` allocate `eltype(r)` to be `promote_type(eltype(a), eltype(b), eltype(c))` (perhaps with an `@inplace` function?). There's plenty of other code that simply duck-types the storage arrays, assuming their types are wide enough to hold the result of the computation, so this doesn't seem to be entirely out of place, but I wanted to make sure I wasn't missing something performance-critical here. 

Happy to work on this PR, since I have to implement some version of this anyway.

jmxpearson:
- Removing the last of restrictions to` AbstractFloat` instead of `Real`.
- Doc fixes.
- Previously inplace functions in `utils.jl` only accepted diagonal dispatch (i.e., all eltypes had to match). This has been relaxed.
- Generics in `generics.jl` now allocate storage for inplace methods with element type that is the promotion of the element types of arguments.

Closes #44.


jmxpearson:
I'm leaving this up for a bit for others to take a look and comment before I merge.

PR #48, opened 2016-09-13
Use new dot syntax for vectorized functions

andreasnoack:


src/pdiagmat.jl
```diff
@@ -44,7 +44,7 @@ end
 ### Algebra
 
 inv(a::PDiagMat) = PDiagMat(a.inv_diag, a.diag)
-logdet(a::PDiagMat) = sum(log(a.diag))
+logdet(a::PDiagMat) = sum(@compat(log.(a.diag)))
```
ararslan: In this case we could also use `sum(log, a.diag)` but I don't think it matters much.

PR #49, opened 2016-09-13
Use the more efficient reduction sum(log, x) instead of sum(log.(x))

andreasnoack:


ararslan:
Heh, I should have waited to merge your previous one. As an aside, congrats on commit \#100!

Issue #50, opened 2016-10-24
Stack overflow error when scaling


crbinz:
```
julia> a = PDMat(eye(3))
PDMats.PDMat{Float64,Array{Float64,2}}(3,[1.0 0.0 0.0; 0.0 1.0 0.0; 0.0 0.0 1.0],Base.LinAlg.Cholesky{Float64,Array{Float64,2}} with factor:
[1.0 0.0 0.0; 0.0 1.0 0.0; 0.0 0.0 1.0])

julia> 3 * a
ERROR: StackOverflowError:
 in *(::PDMats.PDMat{Float64,Array{Float64,2}}, ::Int64) at /Users/cbinz/.julia/v0.5/PDMats/src/generics.jl:24 (repeats 80000 times)
```

This works as expected when the scaling factor is a `Float64`, by the way.

crbinz:
[Here](https://github.com/JuliaStats/PDMats.jl/blob/master/src/pdmat.jl#L37), this will happen any time the argument of `c` is not the same type as `a`.  I think this could just be changed like this: 

```
*{S<:Real, T<:Real}(a::PDMat{S}, c::T) = PDMat(a.mat * c)
```

along with the relevant changes [here](https://github.com/JuliaStats/PDMats.jl/blob/master/src/generics.jl#L24). However, I don't know enough to say whether this is a good idea or not, or what other implications this might have.

crbinz:
Closed by #54

PR #54, opened 2016-12-18
Allow multiplication with scalar of different type

crbinz:
Addresses #50. I got no feedback there, so I just decided to implement it. Open to feedback here, I don't think this should be too controversial...

### DIFF
--- a/test/generics.jl
+++ b/test/generics.jl
@@ -1,0 +1,24 @@
+
+# test operators with pd matrix types
+using PDMats
+using Base.Test
+
+# test scalar multiplication 
+print_with_color(:blue, "Testing scalar multiplication\n")
+pm1 = PDMat(eye(3))
+pm2 = PDiagMat(ones(3))
+pm3 = ScalMat(3,1)
+
+pm1a = PDMat(3.0 .* eye(3))
+pm2a = PDiagMat(3.0 .* ones(3))
+pm3a = ScalMat(3, 3)
+
+pmats = Any[pm1, pm2, pm3]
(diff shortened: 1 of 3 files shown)


ararslan:
This looks good to me. Not sure why we didn't allow this before. Could you add some tests to ensure that this produces the expected behavior?

crbinz:
Done. Besides checking that it didn't result in a stack overflow, I couldn't really think of a better way to check this.

test/generics.jl
```diff
@@ -0,0 +1,24 @@
...
+using Base.Test
+
+# test scalar multiplication 
+print_with_color(:blue, "Testing scalar multiplication\n")
+pm1 = PDMat(eye(3))
+pm2 = PDiagMat(ones(3))
+pm3 = ScalMat(3,1)
+
+pm1a = PDMat(3.*eye(3))
+pm2a = PDiagMat(3.*ones(3))
```
ararslan: Should space out the `.*`, otherwise it's not clear whether it should be dot-broadcast (i.e. `3 .* eye`) or float literal multiplication (i.e. `3. * eye`).

ararslan:
> Besides checking that it didn't result in a stack overflow, I couldn't really think of a better way to check this.

`eltype(A + c) = Base.promote_op(+, eltype(A), typeof(c))`, where `A` is a `PDMat` with one element type and `c` is a scalar with a different type, should suffice.

crbinz:
Do you suggest I replace what's there now with something like that then?

ararslan:
Nah, what you have there is fine.

PR #56, opened 2017-01-17
Add back wrongly deleted tests

andreasnoack:


tkelman:
there was also
```
if VERSION >= v"0.4.2"
    s = speye(Float32, 2, 2)
    @test convert(PDSparseMat{Float64}, PDSparseMat(s)).mat == PDSparseMat(convert(SparseMatrixCSC{Float64}, s)).mat
end
```
at the end

tkelman:
does sparse need wider tolerances?

andreasnoack:
Might be. I can't reproduce the error on my Julia 04 on the Mac. I'll try on a Linux machine.

andreasnoack:
I could reproduce it on Linux. It appears that the `isapprox` tolerance is just wrong on 0.4. It uses the most precise of the machine precisions so this will probably cause a lot of problems when trying to get tests to pass on 0.4.

src/testutils.jl
```diff
@@ -122,7 +122,9 @@ end
 
 function pdtest_logdet(C::AbstractPDMat, Cmat::Matrix, verbose::Int)
     _pdt(verbose, "logdet")
-    @test logdet(C) ≈ logdet(Cmat)
+    # default tolerance in isapprox is different on 0.4. rtol argument can be delted
```
tkelman: deleted

PR #61, opened 2017-04-11
Only define methods for sparse matrices when CHOLMOD is available.

### LINKED ISSUE #43: PDMats not working on Julia 0.4.x compiled without GPL libraries
For licensing issues we run a build of julia excluding GPL libraries. (USE_GPL_LIBS=0)

PDMats seems to be incompatible with this:

```
ERROR: LoadError: LoadError: UndefVarError: CHOLMOD not defined
 in include at ./boot.jl:261
 in include_from_node1 at ./loading.jl:304
 in include at ./boot.jl:261
 in include_from_node1 at ./loading.jl:304
while loading /dynizer/.julia_pkg/v0.4/PDMats/src/chol.jl, in expression starting on line 16
while loading /dynizer/.julia_pkg/v0.4/PDMats/src/PDMats.jl, in expression starting on line 50
```

andreasnoack:
Fixes #43. This will most likely also drop support for 0.4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,10 @@ os:
   - linux
   - osx
 julia:
-  - 0.4
   - 0.5
+  - 0.6
   - nightly
 notifications:
   email: false
-# script:
-#   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-#   - julia --check-bounds=yes -e 'Pkg.clone(pwd()); Pkg.build("PDMats"); Pkg.test("PDMats"; coverage=true)'
 after_success:
   - julia -e 'cd(Pkg.dir("PDMats")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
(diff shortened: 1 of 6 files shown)


ararslan:
The `@static`s shouldn't be necessary at the top level

ararslan:
Looks like this is failing on Julia 0.4 since there's no SparseArrays submodule of Base. While I don't mind dropping support for 0.4 at this point, I wonder if this could be handled more effectively by putting all of the sparse stuff in its own file, then doing
```julia
if isdefined(Base, :SparseArrays) && isdefined(Base.SparseArrays, :CHOLMOD)
    include("sparse.jl")
end
```
in src/PDMats.jl.

aviks:
Bump. Could we get this in please? This breaks about a dozen other packages (via Distributions) when run with a non-GPL julia build.

aviks:
Thanks! @andreasnoack . Appreciate it. Is this ready for a new tag?

andreasnoack:
Yes, I think so. I'll do it as soon as all tests have passed.

PR #65, opened 2017-09-15
Fix most depwarns on 0.7

yuyichao:


test/addition.jl
```diff
@@ -15,18 +15,18 @@ for T in [Float64,Float32]
   pm3 = ScalMat(3,X)
   pm4 = X*I
   #do not test PDSparseMat for Float32 on Julia 0.3 - impossible because of issue #14076
-  pm5 = (T==Float64 || VERSION > v"0.4.2")?PDSparseMat(sparse(M)):PDMat(M)
+  pm5 = (T==Float64 || VERSION > v"0.4.2") ? PDSparseMat(sparse(M)) : PDMat(M)
```
ararslan: 0.4.2 is weirdly specific. You don't have to do it now but we could probably remove that check at some point since the package requires 0.5.

andreasnoack: Yes. I think it would be fine to remove as part of this.

yuyichao: Done and also removed other 0.3/0.4 compat/version checks.

PR #66, opened 2017-10-08
changed import Base.Test to import Test

xiaodaigh:
Base.Test is deprecated in favour of Test

PR #67, opened 2017-10-20
Use Compat.Test instead of Base.Test for Julia v0.7 compatibility

oschulz:
Similar to #66, but compatible to pre-v0.7

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,2 @@
 julia 0.5
-Compat 0.18.0
+Compat 0.33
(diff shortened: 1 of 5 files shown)


oschulz:
Fixes #68 .

oschulz:
Would be nice to tag PDMats after this (or another fix) is merged.

ararslan:
I'm surprised a `using Compat` isn't necessary before the `using Compat.Test`.

oschulz:
Thanks for the quick merge, @ararslan !

ararslan:
Thanks for the contribution 👍

PR #69, opened 2017-11-10
Fix deprecations

Nosferican:


ararslan:
It would be great if you could remove 0.5 from the Travis matrix.

PR #72, opened 2018-02-26
WIP: Updates for Julia 0.7

### LINKED ISSUE #71: Precompilation broken: unknown Base.LAPACK with Julia 0.7
```Base.LAPACK``` has been removed in Julia 0.7. However, ```Base.LinAlg.LAPACK``` still exists, even if it's deprecated.

The precompilation currently fails as ```Base.LAPACK``` cannot be imported on line 40 of PDMats.jl:

https://github.com/JuliaStats/PDMats.jl/blob/a134789404241f24bcf6eb7a397e8f8591f83705/src/PDMats.jl#L40

I would recommend using ```Base.LinAlg.LAPACK``` instead, which works for Julia < 0.7 and 0.7.

ararslan:
This PR drops support for Julia versions before the linear algebra functionality was moved to the stdlib and rewrites the `A_mul_B`-style calls to 0.7-style.

Work in progress because I'm not 100% certain I got all of the in-place calls right.

Fixes #71.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 language: julia
 os:
   - linux
-  - osx
 julia:
-  - 0.6
   - nightly
 notifications:
   email: false
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,1 @@
-julia 0.6
-Compat 0.33
+julia 0.7.0-DEV.3449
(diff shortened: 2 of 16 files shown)


ararslan:
I think things are mostly okay now save for the in-place linear algebra stuff. @andreasnoack, do you have any sagely guidance for how those should be written?

ararslan:
Closing in favor of #73. Forgot Doug was going to take this on.

PR #73, opened 2018-03-01
Changes for Julia v0.7

dmbates:
This is a work in progress on which I would appreciate comments.  I'm not sure what to do about sparse matrices.  At present in v0.7.0-DEV the CHOLMOD types and methods are in the SuiteSparse stdlib package but I don't know if they will stay there, because of the [L]GPL issues.  For the time being I have disabled the sparse parts.

As a parallel project I am working on switching the documentation to the Documenter package.  I see that a few vestiges have crept in here.

What I want to get done is

- [ ] `Base.LinAlg` -> `LinearAlgebra`
- [ ] method definitions for foreign generics use fully qualified names
- [ ] `full()` -> `Matrix()`
- [ ] `A[ct]_mul_B[ct]!` -> `mul!`; same for ldiv and rdiv

### DIFF
--- a/src/PDMats.jl
+++ b/src/PDMats.jl
@@ -3,7 +3,7 @@ __precompile__()
 module PDMats
 
     using Compat
-    using LinearAlgebra, SparseArrays
+    using IterativeEigensolvers, LinearAlgebra, SparseArrays, SuiteSparse
 
     import Base: +, *, \, /, ==, convert, inv, Matrix
     import LinearAlgebra: logdet, diag, eigmax, eigmin
@@ -37,14 +37,12 @@ module PDMats
         Xt_invA_X,
         test_pdmat
 
-#    import Base.BLAS: nrm2, axpy!, gemv!, gemm, gemm!, trmv, trmv!, trmm, trmm!
-#    import Base.LAPACK: trtrs!
-    import LinearAlgebra: A_ldiv_B!, A_mul_B!, A_mul_Bc!, A_rdiv_B!, A_rdiv_Bc!, Ac_ldiv_B!
-
 
     # The abstract base type
 
     abstract type AbstractPDMat{T<:Real} end
+
+    const HAVE_CHOLMOD = isdefined(SuiteSparse, :CHOLMOD)
 
     # source files
 
@@ -52,11 +50,11 @@ module PDMats
     include("utils.jl")
 
     include("pdmat.jl")
-#=
-    if isdefined(Base.SparseArrays, :CHOLMOD)
+
+    if HAVE_CHOLMOD
         include("pdsparsemat.jl")
     end
-=#
+
     include("pdiagmat.jl")
     include("scalmat.jl")
 
--- a/src/PDMats.jl
+++ b/src/PDMats.jl
@@ -6,7 +6,7 @@ module PDMats
     using LinearAlgebra, SparseArrays
 
     import Base: +, *, \, /, ==, convert, inv, Matrix
-    import LinearAlgebra: logdet, diag, diagm, eigmax, eigmin
+    import LinearAlgebra: logdet, diag, eigmax, eigmin
 
     export
         # Types
--- a/src/PDMats.jl
+++ b/src/PDMats.jl
@@ -3,9 +3,10 @@ __precompile__()
 module PDMats
 
     using Compat
+    using LinearAlgebra, SparseArrays
 
-    import Base: +, *, \, /, ==
-    import Base: full, logdet, inv, diag, diagm, eigmax, eigmin, convert
+    import Base: +, *, \, /, ==, convert, inv, Matrix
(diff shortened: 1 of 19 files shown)


src/PDMats.jl
```diff
@@ -36,9 +37,9 @@ module PDMats
         Xt_invA_X,
         test_pdmat
 
-    import Base.BLAS: nrm2, axpy!, gemv!, gemm, gemm!, trmv, trmv!, trmm, trmm!
-    import Base.LAPACK: trtrs!
-    import Base.LinAlg: A_ldiv_B!, A_mul_B!, A_mul_Bc!, A_rdiv_B!, A_rdiv_Bc!, Ac_ldiv_B!, Cholesky
+#    import Base.BLAS: nrm2, axpy!, gemv!, gemm, gemm!, trmv, trmv!, trmm, trmm!
+#    import Base.LAPACK: trtrs!
+    import LinearAlgebra: A_ldiv_B!, A_mul_B!, A_mul_Bc!, A_rdiv_B!, A_rdiv_Bc!, Ac_ldiv_B!
```
ararslan: We'll want to get rid of these imports, since these are deprecated.

src/addition.jl
```diff
@@ -1,36 +1,36 @@
...
 # between pdmat and pdmat
 
-+(a::PDMat, b::AbstractPDMat) = PDMat(a.mat + full(b))
-+(a::PDiagMat, b::AbstractPDMat) = PDMat(_adddiag!(full(b), a.diag))
-+(a::ScalMat, b::AbstractPDMat) = PDMat(_adddiag!(full(b), a.value))
-if isdefined(Base.SparseArrays, :CHOLMOD)
++(a::PDMat, b::AbstractPDMat) = PDMat(a.mat + Matrix(b))
++(a::PDiagMat, b::AbstractPDMat) = PDMat(_adddiag!(Matrix(b), a.diag))
++(a::ScalMat, b::AbstractPDMat) = PDMat(_adddiag!(Matrix(b), a.value))
+if isdefined(SparseArrays, :CHOLMOD)
```
ararslan: This will actually always be `false`, since the sparse linear algebra functionality was moved to the stdlib package SuiteSparse rather than SparseArrays. You could do
```julia
const HAVE_CHOLMOD = isdefined(SuiteSparse, :CHOLMOD)

if HAVE_CHOLMOD
    # methods and whatnot
end
```

src/pdsparsemat.jl
```diff
@@ -84,22 +84,22 @@ end
...
     z = x*sparse(chol_lower(a.chol))
-    A_mul_Bt(z, z)
+    z * z'
 end
 
 
 function Xt_A_X(a::PDSparseMat, x::StridedMatrix)
-    z = At_mul_B(x, sparse(chol_lower(a.chol)))
-    A_mul_Bt(z, z)
+    z = x' * sparse(chol_lower(a.chol))
```
ararslan: Note that the direct translation from the `A_op_B` style would be to use `transpose(x)`, since `x'` is the adjoint. You were using `transpose` in src/pdmat.jl, probably best to be consistent.

src/pdsparsemat.jl
```diff
@@ -84,22 +84,22 @@ end
...
+    z * z'
 end
 
 
 function X_invA_Xt(a::PDSparseMat, x::StridedMatrix)
-    z = a \ x'
+    z = a.chol \ collect(x')
```
ararslan: Why does `x` have to be materialized? Is there no appropriate solver method for whatever type `a.chol` is?

dmbates: No, there isn't an appropriate solver.

src/pdsparsemat.jl
dmbates: Thanks for catching that.  I just pushed a commit with those changes.  I hope it is ready to go now.

src/PDMats.jl
```diff
@@ -36,24 +36,24 @@ module PDMats
...
 
     # The abstract base type
 
     abstract type AbstractPDMat{T<:Real} end
 
+    const HAVE_CHOLMOD = isdefined(SuiteSparse, :CHOLMOD)
+
     # source files
 
-    include("chol.jl")   # make Cholesky compatible with both Julia 0.3 & 0.4
```
ararslan: Come to think of it, if this was only included for compatibility with Julia 0.3 and 0.4, would it be possible to remove it entirely and just use `Cholesky` directly? More of a general wondering though, not something you need to take on here or anything.

ararslan:
This is looking fantastic. The move from `full` to `Matrix` will require a deprecation, but aside from that I think we're good here. Thanks so much for taking this on!

ararslan:
I added the deprecation for `full`. Once CI finishes I think we should merge and tag.

PR #75, opened 2018-03-20
Use `Compat.undef` instead of `uninitialized`

dmbates:
Cut down on the noise when testing on v0.7

### DIFF
--- a/src/generics.jl
+++ b/src/generics.jl
@@ -83,7 +83,7 @@ the quadratic form is applied column-wise.
 """
 function quad(a::AbstractPDMat{T}, x::StridedMatrix{S}) where {T<:Real, S<:Real}
     @check_argdims dim(a) == size(x, 1)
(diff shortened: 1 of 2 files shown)


src/generics.jl
```diff
@@ -83,7 +83,7 @@ the quadratic form is applied column-wise.
 """
 function quad(a::AbstractPDMat{T}, x::StridedMatrix{S}) where {T<:Real, S<:Real}
     @check_argdims dim(a) == size(x, 1)
-    quad!(Array{promote_type(T, S)}(uninitialized, size(x,2)), a, x)
+    quad!(Array{promote_type(T, S)}(Compat.undef, size(x,2)), a, x)
```
ararslan: This can actually just be `undef`, since `undef` is exported from Compat and we have a `using Compat`. The version of Compat in REQUIRE should be set to 0.61.0 for this as well.

PR #78, opened 2018-06-25
Use SuiteSparse and Arpack packages

dmbates:


REQUIRE
```diff
@@ -1,2 +1,4 @@
 julia 0.7.0-DEV.5190 # for cholesky function
 Compat 0.61.0
+SuiteSparse
```
ararslan: So currently we're in a funny situation where SuiteSparse is a registered package, yet we don't want to use that, we want to use SuiteSparse from the stdlib. So `using SuiteSparse` is good, but it should be removed from REQUIRE.

Element-126:
You beat me! I was in the process of filing an issue about IterativeEigensolvers breaking precompilation on 0.7 beta, when I hit F5 and saw your pull request. Great job guys :+1:

ararslan:
Thanks Doug!

PR #81, opened 2018-07-10
cholesky accessor

rofinn:
Adds support for a `cholesky` accessor method where appropriate. Related to JuliaStats/Distributions.jl#731.

~~NOTE: Tests currently fail locally because of JuliaLinearAlgebra/Arpack.jl#12~~

### DIFF
--- a/src/pdmat.jl
+++ b/src/pdmat.jl
@@ -30,7 +30,7 @@ Base.convert(::Type{AbstractArray{T}}, a::PDMat{T}) where {T<:Real} = a
 dim(a::PDMat) = a.dim
 Base.Matrix(a::PDMat) = copy(a.mat)
 LinearAlgebra.diag(a::PDMat) = diag(a.mat)
-LinearAlgebra.cholesky(a::PDMat) = copy(a.chol)
+LinearAlgebra.cholesky(a::PDMat) = a.chol
 
 
 ### Arithmetics
--- a/src/pdmat.jl
+++ b/src/pdmat.jl
@@ -30,6 +30,7 @@ Base.convert(::Type{AbstractArray{T}}, a::PDMat{T}) where {T<:Real} = a
 dim(a::PDMat) = a.dim
 Base.Matrix(a::PDMat) = copy(a.mat)
 LinearAlgebra.diag(a::PDMat) = diag(a.mat)
+LinearAlgebra.cholesky(a::PDMat) = copy(a.chol)
 
 
 ### Arithmetics
--- a/src/pdsparsemat.jl
+++ b/src/pdsparsemat.jl
@@ -30,6 +30,7 @@ Base.convert(::Type{PDSparseMat{T}}, a::PDSparseMat) where {T<:Real} = PDSparseM
 dim(a::PDSparseMat) = a.dim
 Base.Matrix(a::PDSparseMat) = Matrix(a.mat)
 LinearAlgebra.diag(a::PDSparseMat) = diag(a.mat)
+LinearAlgebra.cholesky(a::PDSparseMat) = copy(a.chol)
 
 
 ### Arithmetics
--- a/src/pdsparsemat.jl
+++ b/src/pdsparsemat.jl
@@ -30,7 +30,7 @@ Base.convert(::Type{PDSparseMat{T}}, a::PDSparseMat) where {T<:Real} = PDSparseM
 dim(a::PDSparseMat) = a.dim
 Base.Matrix(a::PDSparseMat) = Matrix(a.mat)
 LinearAlgebra.diag(a::PDSparseMat) = diag(a.mat)
-LinearAlgebra.cholesky(a::PDSparseMat) = copy(a.chol)
+LinearAlgebra.cholesky(a::PDSparseMat) = a.chol
 
 
 ### Arithmetics
--- a/src/testutils.jl
+++ b/src/testutils.jl
@@ -11,6 +11,7 @@ function test_pdmat(C::AbstractPDMat, Cmat::Matrix;
                     verbose::Int=2,             # the level to display intermediate steps
                     cmat_eq::Bool=false,        # require Cmat and Matrix(C) to be exactly equal
                     t_diag::Bool=true,          # whether to test diag method
+                    t_cholesky::Bool=true,      # whether to test cholesky method
                     t_scale::Bool=true,         # whether to test scaling
                     t_add::Bool=true,           # whether to test pdadd
                     t_logdet::Bool=true,        # whether to test logdet method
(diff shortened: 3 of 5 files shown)


src/testutils.jl
```diff
@@ -96,6 +98,23 @@ function pdtest_diag(C::AbstractPDMat, Cmat::Matrix, cmat_eq::Bool, verbose::Int
...
+        @test cholesky(C).U == cholesky(Cmat).U
+    else
+        @test cholesky(C).U ≈ cholesky(Cmat).U
+    end
+end
+
+function pdtest_cholesky(C::PDSparseMat, Cmat::Matrix, cmat_eq::Bool, verbose::Int)
+    _pdt(verbose, "cholesky")
+    if cmat_eq
+        @test SuiteSparse.CHOLMOD.Sparse(cholesky(C)) == SuiteSparse.CHOLMOD.Sparse(cholesky(Cmat))
```
ararslan: Travis is complaining about this:
```
Error During Test at /home/travis/build/JuliaStats/PDMats.jl/src/testutils.jl:113
  Test threw exception MethodError(SuiteSparse.CHOLMOD.Sparse, (Cholesky{Float64,Array{Float64,2}}([2.0 -1.0 -0.5; -2.0 2.0 -0.75; -1.0 -1.0 2.27761], 'U', 0),), 0xffffffffffffffff)
  Expression: (SuiteSparse.CHOLMOD).Sparse(cholesky(C)) == (SuiteSparse.CHOLMOD).Sparse(cholesky(Cmat))
  MethodError: no method matching SuiteSparse.CHOLMOD.Sparse(::Cholesky{Float64,Array{Float64,2}})
  Closest candidates are:
    SuiteSparse.CHOLMOD.Sparse(::Any) where T<:AbstractArray at abstractarray.jl:22
    SuiteSparse.CHOLMOD.Sparse(!Matched::Ptr{SuiteSparse.CHOLMOD.C_Sparse{Tv<:Union{Complex{Float64}, Float64}}}) where Tv<:Union{Complex{Float64}, Float64} at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v0.7/SuiteSparse/src/cholmod.jl:262
    SuiteSparse.CHOLMOD.Sparse(!Matched::Integer, !Matched::Integer, !Matched::Array{Int64,1}, !Matched::Array{Int64,1}, !Matched::Array{#s608,1} where #s608<:Union{Complex{Float64}, Float64}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v0.7/SuiteSparse/src/cholmod.jl:881
    ...
  Stacktrace:
   [1] SuiteSparse.CHOLMOD.Sparse(::Cholesky{Float64,Array{Float64,2}}) at ./abstractarray.jl:22
   [2] pdtest_cholesky(::PDSparseMat{Float64,SparseMatrixCSC{Float64,Int64}}, ::Array{Float64,2}, ::Bool, ::Int64) at /home/travis/build/JuliaStats/PDMats.jl/src/testutils.jl:113
```

rofinn: Yeah, I might just drop adding a cholesky function for PDSparseMat because I can't get master to build locally (issues with Arpack -> BinaryProvider -> libgfortran).

ararslan: You can use a nightly binary for Julia, since all of our official binaries package libgfortran. That should solve the ARPACK issues. Currently ARPACK doesn't seem to work with source builds of Julia.

andreasnoack: Yes. Please just use the official binary for package development.

rofinn:
Alright, the tests pass locally for me, but the `PDSparseMat` test is kinda gross now.

src/pdmat.jl
```diff
@@ -26,6 +26,7 @@ Base.convert(::Type{AbstractArray{T}}, a::PDMat) where {T<:Real} = convert(PDMat
 dim(a::PDMat) = a.dim
 Base.Matrix(a::PDMat) = copy(a.mat)
 LinearAlgebra.diag(a::PDMat) = diag(a.mat)
+LinearAlgebra.cholesky(a::PDMat) = copy(a.chol)
```
andreasnoack: I'm not sure this one should copy.

nickrobinson251:
How do you feel about adding a method for `PDiagMat` to this PR too? :)

oxinabox:
bump?

src/pdmat.jl
rofinn: I'm more confident that it should be a copy if we're going to support `PDiagMat` which wouldn't be a strict accessor

rofinn:
NOTE: I've added support for `PDiagMat`, but that technically isn't an accessor anymore.

davidanthoff:
And another bump, this would be really nice to have.

davidanthoff:
Actually, maybe also add it for `ScalMat`?

src/pdmat.jl
andreasnoack: We usually do the minimal amount of copying required and allow the aliasing behavior to be different for different types so I still don't think this should copy.

rofinn:
@davidanthoff Should that return `cholesky(Matrix(scalmat))`?

davidanthoff:
I would think so? But @andreasnoack probably knows better. 

Now that I looked a bit more carefully at the package, I'm actually slightly confused. Why isn't `AbstractPDMat` a subtype of `AbstractMatrix`?

src/pdsparsemat.jl
```diff
@@ -30,6 +30,7 @@ Base.convert(::Type{PDSparseMat{T}}, a::PDSparseMat) where {T<:Real} = PDSparseM
 dim(a::PDSparseMat) = a.dim
 Base.Matrix(a::PDSparseMat) = Matrix(a.mat)
 LinearAlgebra.diag(a::PDSparseMat) = diag(a.mat)
+LinearAlgebra.cholesky(a::PDSparseMat) = copy(a.chol)
```
andreasnoack:
```suggestion
LinearAlgebra.cholesky(a::PDSparseMat) = a.chol
```

rofinn:
Thanks @andreasnoack and @davidanthoff!

davidanthoff:
Thank you!

davidanthoff:
Oh, and one more thing: @andreasnoack could you also tag a new release? Thanks :)

johnczito:
Should we say in the README that `cholesky` should be one of the necessary methods for customized subtypes? If not, are folks open to a generic fallback like this?
```
cholesky(A::AbstractPDMat) = cholesky(Matrix(A))
```

davidanthoff:
The more I think about this, the more I think `AbstractPDMat` should just derive from `AbstractMatrix`, in which case I would assume everything would just work automatically. Accessor methods like the one in this PR here then would be performance improvements, but I would assume that the functionality would just be there (saying all of this under the assumption that there is a `cholesky` that works for any `AbstractMatrix`, which I haven't actually checked).

andreasnoack:
So, unfortunetely, this broke support for Julia builds with `USE_GPL_LIBS=0` since `PDSparseMat` is only defined when `SuiteSparse` is available and is now used unconditionally on https://github.com/JuliaStats/PDMats.jl/blob/78cb5065a44e15e20adfdfa79d3e0991707cde04/src/PDMats.jl#L42. @rofinn Would you be able to submit a fix? Eventually, it would also be good to test this but it probably requires that we build Julia from source which I don't think is feasible.

rofinn:
I'm not sure I follow. How is it being used unconditionally now? Oh, is this because testutils is in src/ rather than test/?

andreasnoack:
The problem is that the struct `PDSparseMat` isn't defined when `USE_GPL_LIBS=0` but it is referenced in https://github.com/JuliaStats/PDMats.jl/blob/78cb5065a44e15e20adfdfa79d3e0991707cde04/src/testutils.jl#L33

PR #86, opened 2018-12-01
Remove input eltype restriction to inner constructor of PDiagMat

mohamed82008:
This PR removes the restriction on the `eltype`s of the inputs to the `PDiagMat` constructor. This allows conversion to take place inside with compatible types as opposed to just throwing an error. This was causing trouble in [Turing.jl](https://github.com/TuringLang/Turing.jl).


mohamed82008:
Bump

PR #89, opened 2019-04-11
Relax some strided vec or mat to abstract vec or mat

mohamed82008:
I relaxed some type constraints to partly enable the use of `Tracker.TrackedArray` with `Distributions.logpdf` and `Distributions.MvNormal`.

Relevant: https://github.com/TuringLang/Turing.jl/pull/743


mohamed82008:
Reference for the Cholesky gradient: https://homepages.inf.ed.ac.uk/imurray2/pub/16choldiff/choldiff.pdf

ararslan:
Seems reasonable to me, but I fully admit I don't know why these were limited to strided arrays in the first place, so perhaps there is a reason the signatures are how they are.

Issue #94, opened 2019-05-09
More Stable `invquad(A::PDMat, z::AbstractVector)`


bayesthm:
For A a `PDMat` and z a `Vector`, would it not be more numerically stable to compute `invquad(A, z)`  as 
```
zhalf  = A.chol.L \ z
q = dot(zhalf, zhalf)
```
rather than the current:

https://github.com/JuliaStats/PDMats.jl/blob/3c74b56dfd91986967b5bec20fa0ed06be6c201d/src/pdmat.jl#L72

In some experiments my suggested way of computing `invquad(A,z)` more reliably gives positive values (as it always should!) when A and z contain very large values and A is close to singular.  This is important when computing `logpdf()` of a multivariate normal distribution at fairly extreme values using the Distributions.jl package, which appears to use the PDMats `invquad()` function.  

If I understand the source code correctly this is already how the computation is handled in the case of z being a matrix:

https://github.com/JuliaStats/PDMats.jl/blob/3c74b56dfd91986967b5bec20fa0ed06be6c201d/src/pdmat.jl#L110

An analogous suggestion could also pertain to the computing of `quad(A, z)`, though the numerical stability properties there are probably less of a problem.  I'm running Julia 1.1.0 and PDMats 0.9.7.

andreasnoack:
It would be more stable and faster as well. Would you be able to prepare a PR?

bayesthm:
Yes, I will do so.

devmotion:
Fixed by https://github.com/JuliaStats/PDMats.jl/pull/95.

PR #95, opened 2019-06-03
more stable quad and invquad with vectors

bayesthm:
More stable versions of `quad` and `invquad` for the case of `(a::PDMat, x::AbstractVector)`.
This addresses issue #94 

`quad` uses the fact that: 
x'*a*x = x'*U'*U*x = (U*x)'*(U*x)

`invquad` uses the fact that:
x'*inv(a)*x = x'*inv(L*L')*x
                 = x'*inv(L')*inv(L)*x
                 = x'*inv(L)'*inv(L)*x
                 = (L \ x)'*(L \ x)

### DIFF
--- a/src/pdmat.jl
+++ b/src/pdmat.jl
@@ -68,8 +68,15 @@ end
 
 ### quadratic forms
 
-quad(a::PDMat, x::AbstractVector) = dot(x, a * x)
-invquad(a::PDMat, x::AbstractVector) = dot(x, a \ x)
+function quad(a::PDMat, x::AbstractVector)
+    z = a.chol.U * x
+    dot(z, z)
+end
+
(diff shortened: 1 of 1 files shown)


src/pdmat.jl
```diff
@@ -68,8 +68,15 @@ end
 
 ### quadratic forms
 
-quad(a::PDMat, x::AbstractVector) = dot(x, a * x)
-invquad(a::PDMat, x::AbstractVector) = dot(x, a \ x)
+function quad(a::PDMat, x::AbstractVector)
+    z = a.chol.U * x
+    dot(z, z)
```
ararslan: You might even be able to speed this up a bit using `sum(abs2, z)`. I've found that is often faster than `dot(z, z)`.

PR #97, opened 2019-06-16
make type abstract

adam-r-kowalski:
I want to use this package with Flux Tracked Vectors so I loosened the requirement from a Vector to an Abstract Vector.


charleskawczynski:
Can we merge this? I'm running into this also.

charleskawczynski:
Should we also convert the `::Vector` types (in `utils.jl`) while we're at it? How about `::Matrix` to `AbstractMatrix`?

andreasnoack:
> Should we also convert the ::Vector types (in utils.jl) while we're at it? How about ::Matrix to AbstractMatrix?

Probably but sometimes ambiguities show up when you try to make such changes

PR #101, opened 2019-07-29
Don't do anything if converting to AbstractArray with correct eltype

kleinschmidt:
This PR adds a no-op `convert` method for when convert to an `AbstractArray`
with the same eltype as the PDMat.  This is a workaround to support things like
static arrays wrapped in PDMats.


ararslan:
A test would be nice, perhaps something as simple as
```julia
X = PDMat((Y->Y'Y)(randn(Float32, 4, 4)))
@test convert(AbstractArray{Float32}, X) === X
```

PR #103, opened 2019-08-31
more kron

johnczito:
Continuing the work of #100, this adds `kron` methods for `PDiagMat` and `ScalMat`. It also provides a generic fallback that works on any pair of (possibly different) `AbstractPDMat` subtypes.


johnczito:
Bump.

Issue #40, opened 2016-02-11
PDMats implementing AbstractArray interface


KrisDM:
Working with PDMats, it made me wonder why AbstractPDMat doesn't derive from AbstractMatrix? Given the class is immutable, we'd only have to implement getindex and size I think.

theogf:
It's an old issue but any comments on this?

davidanthoff:
@andreasnoack what is your take on this? Worth opening a PR?

johnczito:
Bump. I think this would be helpful.

johnczito:
This would be helpful. I'm currently working on `kron` for PDMats, and this would provide a clean [fallback](https://github.com/JuliaLang/julia/blob/f928246ec3fc5e12ef995388324a8fdcd59cb944/stdlib/LinearAlgebra/src/dense.jl#L384) for any pair of arg types. Then as [you](https://github.com/JuliaStats/PDMats.jl/pull/81#issuecomment-520103299) said, anything else would just be a performance improvement.

devmotion:
Fixed by https://github.com/JuliaStats/PDMats.jl/pull/105.

PR #105, opened 2019-09-06
Making AbstractPDMat inherit from AbstractMatrix

theogf:
Following issue #40 
Implemented `getindex` for all types for a nicer display of the PDMats

### DIFF
--- a/src/pdiagmat.jl
+++ b/src/pdiagmat.jl
@@ -27,6 +27,8 @@ dim(a::PDiagMat) = a.dim
 Base.Matrix(a::PDiagMat) = Matrix(Diagonal(a.diag))
 LinearAlgebra.diag(a::PDiagMat) = copy(a.diag)
 LinearAlgebra.cholesky(a::PDiagMat) = cholesky(Diagonal(a.diag))
+Base.getindex(a::PDiagMat{T},i::Integer) where {T} = i%a.dim == (i÷a.dim+1) ? a.diag[i%a.dim] : zero(0)
+Base.getindex(a::PDiagMat{T},i::Integer,j::Integer) where {T} = i == j ? a.diag[i] : zero(T)
 
 ### Arithmetics
 
--- a/src/pdiagmat.jl
+++ b/src/pdiagmat.jl
@@ -27,6 +27,10 @@ dim(a::PDiagMat) = a.dim
 Base.Matrix(a::PDiagMat) = Matrix(Diagonal(a.diag))
 LinearAlgebra.diag(a::PDiagMat) = copy(a.diag)
 LinearAlgebra.cholesky(a::PDiagMat) = cholesky(Diagonal(a.diag))
+
+### Inheriting from AbstractMatrix
+
+Base.size(a::PDiagMat) = (a.dim,a.dim)
 Base.getindex(a::PDiagMat{T},i::Integer) where {T} = i%a.dim == (i÷a.dim+1) ? a.diag[i%a.dim] : zero(T)
 Base.getindex(a::PDiagMat{T},i::Integer,j::Integer) where {T} = i == j ? a.diag[i] : zero(T)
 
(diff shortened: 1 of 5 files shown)


theogf:
So following https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array-1
I realized I should implement `size` and `setindex!`.
But what do you think `setindex!` should do? : 
 - Should it just return an error (we assume all `AbstractPDMat` are entirely immutable) ? 
 - Should it allow to change diagonal elements (recomputing Cholesky afterwards) ? 
 - Should it allow to change off-diagonal elements and automatically symmetrize the matrix accordingly (recomputing Cholesky afterwards) ?

jlumpe:
I thin you can leave `setindex!` undefined for non-mutable arrays. `size` should be implemented for `AbstractArray`s.

theogf:
I added `size` and left out `setindex!`, any more details I should take care?
Note that the advantages of inheriting from `AbstractMatrix` also involves having fallback algebraic operations (i.e `Base.:-(PDMat,Matrix)` etc), and a nicer display.

theogf:
Bump

theogf:
Bump

src/pdiagmat.jl
```diff
@@ -41,8 +47,10 @@ function pdadd!(r::Matrix, a::Matrix, b::PDiagMat, c)
 end
 
 *(a::PDiagMat, c::T) where {T<:Real} = PDiagMat(a.diag * c)
-*(a::PDiagMat, x::AbstractVecOrMat) = a.diag .* x
+*(a::PDiagMat, x::AbstractVector) = a.diag .* x
+*(a::PDiagMat, x::AbstractMatrix) = a.diag .* x
 \(a::PDiagMat, x::AbstractVecOrMat) = a.inv_diag .* x
+/(x::AbstractVecOrMat, a::PDiagMat) = a.inv_diag .* x
```
andreasnoack, line 53: Why were these `/` methods added? They seem wrong. E.g.
```julia
julia> randn(3)/randn(3,3)
ERROR: DimensionMismatch("Both inputs should have the same number of columns")
Stacktrace:
 [1] /(A::Vector{Float64}, B::Matrix{Float64})
   @ LinearAlgebra ~/julia/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:1143
 [2] top-level scope
   @ REPL[2]:1
   ```

Comment (theogf) [reply src/pdiagmat.jl:53]: > Why were these / methods added?
Because they were missing :)

But you're right, the current definition for `PDiagMat` is wrong
To be fair it's also wrong for `\` : 

```
julia> x = rand(3)
3-element Array{Float64,1}:
 0.7725835504446747
 0.8135553481482751
 0.2409921677507807

julia> x' isa AbstractVecOrMat
true

julia> rand(3, 3) \ x'
ERROR: DimensionMismatch("B has leading dimension 1, but needs 3")
Stacktrace:
 [1] getrs!(::Char, ::Array{Float64,2}, ::Array{Int64,1}, ::Array{Float64,2}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/lapack.jl:996
 [2] ldiv! at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/lu.jl:391 [inlined]
 [3] \(::LU{Float64,Array{Float64,2}}, ::Adjoint{Float64,Array{Float64,1}}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/factorization.jl:101
 [4] \(::Array{Float64,2}, ::Adjoint{Float64,Array{Float64,1}}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/generic.jl:1116
 [5] top-level scope at REPL[18]:1

julia> 
```

Comment (theogf) [reply src/pdiagmat.jl:53]: Should it be changed to:
```suggestion
/(x::AbstractMatrix, a::PDiagMat) = x .* a.inv_diag'
```

andreasnoack: Ah. We allow the 1x1 case. A bit annoying that we allow that but it's too late to change.

oschulz:
Thanks for this @theogf !

PR #107, opened 2019-11-23
Use Upper Triangular Cholesky to avoid type instability (closes #35)

### LINKED ISSUE #35: Use upper triangular Cholesky
Following up on the discussion in #34 it might be worth using the upper triangular. I shouldn't matter which one is used, but for type stability reasons we don't have an option for the lower version anymore on master.

jebej:
Is this change wanted at this point?


jebej:
bump

PR #111, opened 2020-03-02
More strided -> abstract

johnczito:
As was mentioned in #89 or in #90, it's not clear why operations like `\` require `Strided` vectors or matrices, and it's causing problems in `Distributions.jl`. Density evaluations for `MatrixNormal` and `MatrixTDist` involve things like `S \ A'`, but the `Strided` requirement means you get an error because `A'` is `Adjoint`.


johnczito:
(@matbesancon, @ararslan)

bump

matbesancon:
thanks, merging at the end of the day if no concern raised

PR #115, opened 2020-05-14
Remove eigmin and eigmax definitions for PDSparseMat and thereby

andreasnoack:
stop depending on Arpack. The Arpack dependency sometimes causes some problems on some platforms and since Distributions depend on PDMats it means that most users end up depending on Arpack even though they never compute `eigmax` or `eigmin` of a sparse matrix even indirectly.

### DIFF
--- a/src/pdsparsemat.jl
+++ b/src/pdsparsemat.jl
@@ -1,5 +1,3 @@
-using Arpack
-
 """
 Sparse positive definite matrix together with a Cholesky factorization object.
 """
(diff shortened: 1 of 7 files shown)


andreasnoack:
I've now also enabled testing on the latest Julia release and dropped support for Julia 0.7 to avoid increasing the number of tested versions.

src/pdsparsemat.jl
```diff
@@ -50,9 +48,6 @@ end
 
 Base.inv(a::PDSparseMat{T}) where {T<:Real} = PDMat( a\eye(T,a.dim) )
 LinearAlgebra.logdet(a::PDSparseMat) = logdet(a.chol)
-LinearAlgebra.eigmax(a::PDSparseMat{T}) where {T<:Real} = convert(T,eigs(convert(SparseMatrixCSC{Float64,Int},a.mat), which=:LM, nev=1, ritzvec=false)[1][1]) #to avoid type instability issues in eigs, see e.g., julia issue #13929
```
ararslan, line 53: Are there fallback definitions somewhere that will produce the right thing for a `PDSparseMat`?

andreasnoack: No. The PR is removing functionality based on a trade off between how useful this method is (I don’t think it is) and how costly it is to have the binary dependency for all downstream packages

ararslan: Okay. I don't have an opinion here so I trust your judgement on this as long as we make this a breaking release.

Issue #98, opened 2019-07-03
Cannot construct PDMat from `Matrix{Int}`


nickrobinson251:
The Cholesky will have eltype `Float64` different from the than the original matrix eltype `Int64`, and the constructor does not allow that

```julia
julia> PDMat([1 0; 0 1])
ERROR: MethodError: no method matching PDMat{Int64,Array{Int64,2}}(::Int64, ::Array{Int64,2}, ::Cholesky{Float64,Array{Float64,2}})
Closest candidates are:
  PDMat{Int64,Array{Int64,2}}(::Int64, ::AbstractArray{T,2}, ::Cholesky{T,S}) where {T, S} at /Users/nick/.julia/packages/PDMats/AObTs/src/pdmat.jl:9
Stacktrace:
 [1] PDMat(::Array{Int64,2}, ::Cholesky{Float64,Array{Float64,2}}) at /Users/nick/.julia/packages/PDMats/AObTs/src/pdmat.jl:16
 [2] PDMat(::Array{Int64,2}) at /Users/nick/.julia/packages/PDMats/AObTs/src/pdmat.jl:19
 [3] top-level scope at none:0
```

Came up at https://github.com/JuliaStats/Distributions.jl/pull/919#discussion_r299893423

PR #117, opened 2020-08-12
PDMat: convert Matrix type to the same Cholesky type

mossr:
This addresses #98 where the original Matrix type (e.g., `Matrix{Int}`) can differ from the type returned by `Cholesky` (e.g., `Matrix{Float64}`). Therefore, we simply convert the Matrix type to the same paremetric types already converted by `Cholesky`.

Previously, this failed:
```julia
PDMat([1 0; 0 1])
```

This also addresses https://github.com/JuliaStats/Distributions.jl/issues/1004 (where my original problem started).


andreasnoack:
Could you please add the example as a test? Then I think this one should be good to go.

mossr:
@andreasnoack Good idea—I've since added this example in `pdmtypes.jl`

Issue #121, opened 2020-11-24
Shouldn't isposdef return true for AbstractPDMats?


oschulz:
Currently, we get

```
julia> isposdef(PDMat([1.0 0.0; 0.0 1.0]))
ERROR: MethodError: no method matching cholesky(::Hermitian{Float64,PDMat{Float64,Array{Float64,2}}}; check=false)

julia> isposdef(PDiagMat([1.0, 1.0]))
ERROR: MethodError: no method matching cholesky(::Hermitian{Float64,PDiagMat{Float64,Array{Float64,1}}}; check=false)

julia> isposdef(ScalMat(2, 3.0))
ERROR: MethodError: no method matching cholesky(::Hermitian{Float64,ScalMat{Float64}}; check=false)
```

Shouldn't we define

```julia
LinearAlgebra.isposdef(::AbstractPDMat) = true
```

since `AbstractPDMat`s are positive definite by definition?

oschulz:
I guess we'll probably want to define

```julia
LinearAlgebra.ishermitian(::AbstractPDMat) = true
```

as well, to do that in O(1), since the default implementation will actually test it?

Issue #118, opened 2020-09-17
PDMats.isposdef fails on PDMat type


imLew:
Using version `[90014a1f] PDMats v0.10.0` of PDMats when `PDMats.isposdef` called on a variable of type `PDMat` it raises a MethodError.
To reproduce
```
julia> s = [1. 0; 0 2]
2×2 Array{Float64,2}:
 1.0  0.0
 0.0  2.0

julia> PDMats.isposdef(s)
true

julia> PDMats.isposdef(PDMat(s))
ERROR: MethodError: no method matching cholesky(::LinearAlgebra.Hermitian{Float64,PDMat{Float64,Array{Float64,2}}}; check=false)
```

Possibly related to #117

PR #123, opened 2021-01-05
Add defaults for `isposdef` and `ishermitian`

devmotion:
This PR adds generic definitions of `LinearAlgebra.isposdef` and `LinearAlgebra.ishermitian`. It also removes `Base.eltype` and `Base.ndims` definitions which are not needed anymore since `AbstractPDMat{T} <: AbstractMatrix{T}`.

Fixes https://github.com/JuliaStats/PDMats.jl/issues/121. Fixes https://github.com/JuliaStats/PDMats.jl/issues/118.

I am not completely sure what's the best way to test these definitions (in particular `ishermitian`) - maybe I should add at least the failing examples for `isposdef`?

### DIFF
--- a/test/generics.jl
+++ b/test/generics.jl
@@ -23,7 +23,7 @@ for i in 1:length(pmats)
     @test Matrix(pmats[i] * 3)   == Matrix(pmatsa[i])
 end
 
-# issue #123
+# issue #121
 @test isposdef(PDMat([1.0 0.0; 0.0 1.0]))
 @test isposdef(PDiagMat([1.0, 1.0]))
 @test isposdef(ScalMat(2, 3.0))
--- a/test/generics.jl
+++ b/test/generics.jl
@@ -22,3 +22,8 @@ for i in 1:length(pmats)
     @test Matrix(3 * pmats[i])   == Matrix(pmatsa[i])
     @test Matrix(pmats[i] * 3)   == Matrix(pmatsa[i])
 end
+
+# issue #123
+@test isposdef(PDMat([1.0 0.0; 0.0 1.0]))
+@test isposdef(PDiagMat([1.0, 1.0]))
+@test isposdef(ScalMat(2, 3.0))
--- a/src/generics.jl
+++ b/src/generics.jl
@@ -2,9 +2,6 @@
 
 ## Basic functions
 
-Base.eltype(a::AbstractPDMat{T}) where {T<:Real} = T
-Base.eltype(::Type{AbstractPDMat{T}}) where {T<:Real} = T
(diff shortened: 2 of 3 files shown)


andreasnoack:
> maybe I should add at least the failing examples for isposdef?

Which are they? I think it would be good to add trivial tests. I'm not sure how to test anything non-trivial here since you can't construct the instances if the matrix isn't Hermitian and positive definite.

andreasnoack:
Ah. Yes, it would probably be good to add them.

test/generics.jl
```diff
@@ -22,3 +22,8 @@ for i in 1:length(pmats)
     @test Matrix(3 * pmats[i])   == Matrix(pmatsa[i])
     @test Matrix(pmats[i] * 3)   == Matrix(pmatsa[i])
 end
+
+# issue #123
+@test isposdef(PDMat([1.0 0.0; 0.0 1.0]))
+@test isposdef(PDiagMat([1.0, 1.0]))
+@test isposdef(ScalMat(2, 3.0))
```
palday, line 29: Can you add the same trivial tests for `ishermitian`?

test/generics.jl
```diff
@@ -22,3 +22,8 @@ for i in 1:length(pmats)
     @test Matrix(3 * pmats[i])   == Matrix(pmatsa[i])
     @test Matrix(pmats[i] * 3)   == Matrix(pmatsa[i])
 end
+
+# issue #123
```
devmotion, line 26:
```suggestion
# issue #121
```

test/generics.jl
devmotion: Only `isposdef` was broken, so I only added tests for these failing examples. `ishermitian` works fine for `PDMat`, `PDiagMat`, and `ScalMat` already without this PR, it is just more efficient since it does not have to check all entries explicitly with this PR. So the actual change in this PR can't be tested.

devmotion:
Bump 🙂

palday:
LGTM, @andreasnoack ?

palday (approved):

andreasnoack:
What is the thinking about the case PDiagMat([0.0, 0.0])? I also thinking about the consequences of allowing the zero matrix for PDMat since I have usecase where it would be handy.

devmotion:
My impression was that PDMats (so far) is not designed to work with positive semi-definite matrices, and therefore https://github.com/invenia/PDMatsExtras.jl was created. But maybe it would still be better to define `isposdef` only for PDMat, PDiagMat and ScalMat but not AbstractPDMat? It would also avoid the incorrect default `isposdef(::PSDMat) = true`.

Issue #122, opened 2020-12-01
mul! returns invalid result for PDiagMat


oschulz:
`LinearAlgebra.mul!` returns an invalid result for `PDiagMat`. Given

```julia
using LinearAlgebra, PDMats

A = PDiagMat([0.25, 0.75])
x = [0.0, 1.0]
y = similar(x)
```

We get

```julia
A * x == [0.0, 0.75]
```
as expected, but `mul!` gives

```julia
mul!(y, A, x) == [0.0, 0.0]
```

However,

```julia
mul!(y, Matrix(A), x) == [0.0, 0.75]
```

returns the correct result.

For some reason, `LinearAlgebra.generic_matvecmul!` doesn't seem to work correctly for `PDiagMat`.

Tested with Julia version v1.5.3 on Linux x86_64.

CC @vindex10, who noticed incorrect results caused by this in a actual use case.

oschulz:
Thanks!

PR #124, opened 2021-01-05
Fix linear indexing of `PDiagMat` and `ScalMat`

devmotion:
This PR fixes linear indexing of `PDiagMat` and `ScalMat`. Currently, for the last entry on the diagonal always `0` is returned since `d^2 % d == 0`. This also causes https://github.com/JuliaStats/PDMats.jl/issues/122.

I added some additional tests, both for indexing and for `mul!` (which probably should be implemented more efficiently though).

Fixes https://github.com/JuliaStats/PDMats.jl/issues/122.


andreasnoack:
Why are the `size` methods no longer needed?

devmotion:
They are already defined in https://github.com/JuliaStats/PDMats.jl/blob/d50e35a37d218f0e6ac552512bc4c79b102cc930/src/generics.jl#L8-L9.

PR #131, opened 2021-03-24
Don't promote to `Matrix` with `PDMats` 1-arg constructor

chriselrod:
This should improve support for `StaticArrays`, for example.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,8 @@ SuiteSparse = "4607b0f0-06f3-5cda-b6b1-a6196a1729e9"
 julia = "1"
 
 [extras]
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test"]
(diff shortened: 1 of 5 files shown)


andreasnoack:
Would be good to add a test for this and it should be fine to add StaticArrays as a test dependency.

chriselrod:
I fixed `A::AbstractMatrix / B::PDiagMat`, but I'm not sure what to do about:
```julia
/(x::AbstractVector, a::PDiagMat) = a.inv_diag .* x
```
I'm not sure if this was intentional (probably should've been for `Adjoint`/`Transpose`, although that's covered by `AbstractMatrix`).
Removing it may be considered breaking?

andreasnoack:
Hm. I think they are wrong. I've asked about it in https://github.com/JuliaStats/PDMats.jl/pull/105/files#r600569523. Is it necessary to reach those methods in the tests here?

chriselrod:
I wanted to test a few basic uses (and test that they also return an `SMatrix`). I fixed the incorrect method being used in the tests.

andreasnoack:
@chriselrod It would be great if you could resolve the conflicts here

PR #134, opened 2021-06-11
Fix #133

### LINKED ISSUE #133: Avoid copying when accessing cholesky factors of PDMat
The methods specialized on `PDMat` frequently access the `U` and `L` properties of the stored cholesky factor `chol`. `Cholesky` has an `uplo` parameter that determines whether it stores the upper or lower Cholesky factor. If `chol.uplo=='U'`, then `chol.U` gives an `UpperTriangular` view of the stored factor, while `chol.L` allocates a transposed copy of this matrix. The opposite is true for `chol.L` and `uplo == 'L'`. To avoid unnecessary copies, the `PDMat` methods should check the `uplo` to decide whether to call, e.g. `chol.U` or `chol.L'`

sethaxen:
This PR implements the fix described in #133.

### DIFF
--- a/src/chol.jl
+++ b/src/chol.jl
@@ -1,6 +1,10 @@
 CholType{T,S<:AbstractMatrix} = Cholesky{T,S}
 chol_lower(a::Matrix) = cholesky(a).L
 
+# always use the stored cholesky factor, not a copy
+@inline chol_lower(a::CholType) = a.uplo === 'L' ? a.L : transpose(a.U)
+@inline chol_upper(a::CholType) = a.uplo === 'U' ? a.U : transpose(a.L)
+
 if HAVE_CHOLMOD
     CholTypeSparse{T} = SuiteSparse.CHOLMOD.Factor{T}
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "PDMats"
 uuid = "90014a1f-27ba-587c-ab20-58faa44d9150"
-version = "0.11.0"
+version = "0.11.1"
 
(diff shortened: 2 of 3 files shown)


src/chol.jl
```diff
@@ -1,6 +1,10 @@
 CholType{T,S<:AbstractMatrix} = Cholesky{T,S}
 chol_lower(a::Matrix) = cholesky(a).L
 
+# always use the stored cholesky factor, not a copy
+@inline chol_lower(a::CholType) = a.uplo === 'L' ? a.L : transpose(a.U)
```
andreasnoack, line 5:
```suggestion
chol_lower(a::CholType) = a.uplo === 'L' ? a.L : a.U'
```
I think we should only add annotations if we know they are needed. I'd think the heuristic would handle this one just fine.

src/chol.jl
```diff
@@ -1,6 +1,10 @@
...
+@inline chol_upper(a::CholType) = a.uplo === 'U' ? a.U : transpose(a.L)
```
andreasnoack, line 6:
```suggestion
@inline chol_upper(a::CholType) = a.uplo === 'U' ? a.U : a.L'
```

andreasnoack (commented):
Just a minor comment. Except for that, this should be fine.

sethaxen:
Thanks! Can you approve running the workflow?

andreasnoack:
> Can you approve running the workflow?

Sure. It's the first time I've seen that one.

andreasnoack:
@JuliaRegistrator register

PR #141, opened 2021-11-02
Fix PDiagMat when inv(v) isn't a subtype of v

Wynand:
Currently for any `AbstractVector` `v` where `typeof(inv(v))` isn't a subtype of `typeof(v))` creating a `PDiagMat` will fail

For example, if we use a range such a `0.1:0.1:0.5` to construct a `PDiagMat` we get this error:

```julia
julia> using PDMats
[ Info: Precompiling PDMats [90014a1f-27ba-587c-ab20-58faa44d9150]

julia> PDiagMat(0.1:0.1:0.5)
ERROR: MethodError: Cannot `convert` an object of type Vector{Float64} to an object of type StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}}
Closest candidates are:
  convert(::Type{T}, ::AbstractRange) where T<:AbstractRange at range.jl:148
  convert(::Type{T}, ::LinearAlgebra.Factorization) where T<:AbstractArray at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/factorization.jl:58
  convert(::Type{T}, ::T) where T<:AbstractArray at abstractarray.jl:14
  ...
Stacktrace:
 [1] PDiagMat{Float64, StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}}}(d::Int64, v::StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}}, inv_v::Vector{Float64})
   @ PDMats ~/git-repos/open_source/julia_packages/PDMats.jl/src/pdiagmat.jl:9
 [2] PDiagMat(v::StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}}, inv_v::Vector{Float64})
   @ PDMats ~/git-repos/open_source/julia_packages/PDMats.jl/src/pdiagmat.jl:15
 [3] PDiagMat(v::StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}})
   @ PDMats ~/git-repos/open_source/julia_packages/PDMats.jl/src/pdiagmat.jl:18
 [4] top-level scope
   @ REPL[2]:1

julia> 
``` 

To fix this I've changed [this function](https://github.com/JuliaStats/PDMats.jl/blob/dfcf96c90ade0ad6052855a3e4b3488b3dbe002a/src/pdiagmat.jl#L13) to create a `PDiagMat` with a Union of the two `AbstractVector` types as the types for the two vectors

### DIFF
--- a/src/pdiagmat.jl
+++ b/src/pdiagmat.jl
@@ -1,21 +1,22 @@
 """
 Positive definite diagonal matrix.
 """
-struct PDiagMat{T<:Real,V<:AbstractVector} <: AbstractPDMat{T}
+struct PDiagMat{T<:Real,V<:AbstractVector,I<:AbstractVector} <: AbstractPDMat{T}
     dim::Int
     diag::V
-    inv_diag::V
+    inv_diag::I
 
-    PDiagMat{T,S}(d::Int,v::AbstractVector,inv_v::AbstractVector) where {T,S} =
-        new{T,S}(d,v,inv_v)
+    PDiagMat{T,S,I}(d::Int,v::AbstractVector,inv_v::AbstractVector) where {T,S,I} =
+        new{T,S,I}(d,v,inv_v)
+    PDiagMat{T,S}(d::Int,v::AbstractVector,inv_v::AbstractVector) where {T,S} = PDiagMat{T,S,S}(d,v,inv_v)
 end
 
-function PDiagMat(v::A,inv_v::B) where {A <: AbstractVector, B <: AbstractVector}
+function PDiagMat(v::AbstractVector,inv_v::AbstractVector)
     @check_argdims length(v) == length(inv_v)
-    PDiagMat{eltype(v),Union{A,B}}(length(v), v, inv_v)
+    PDiagMat{eltype(v),typeof(v),typeof(inv_v)}(length(v), v, inv_v)
 end
 
-PDiagMat(v::AbstractVector) = PDiagMat(v, inv.(v))
+PDiagMat(v::AbstractVector) = PDiagMat{eltype(v),typeof(v), typeof(inv.(v))}(length(v), v, inv.(v))
 
 ### Conversion
 Base.convert(::Type{PDiagMat{T}},      a::PDiagMat) where {T<:Real} = PDiagMat(convert(AbstractArray{T}, a.diag))
--- a/src/pdiagmat.jl
+++ b/src/pdiagmat.jl
@@ -67,13 +67,13 @@ end
 *(a::PDiagMat, c::T) where {T<:Real} = PDiagMat(a.diag * c)
 *(a::PDiagMat, x::AbstractVector) = a.diag .* x
 *(a::PDiagMat, x::AbstractMatrix) = a.diag .* x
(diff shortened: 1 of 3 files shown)


src/pdiagmat.jl
```diff
@@ -10,9 +10,9 @@ struct PDiagMat{T<:Real,V<:AbstractVector} <: AbstractPDMat{T}
         new{T,S}(d,v,inv_v)
 end
 
-function PDiagMat(v::AbstractVector,inv_v::AbstractVector)
+function PDiagMat(v::A,inv_v::B) where {A <: AbstractVector, B <: AbstractVector}
     @check_argdims length(v) == length(inv_v)
-    PDiagMat{eltype(v),typeof(v)}(length(v), v, inv_v)
+    PDiagMat{eltype(v),Union{A,B}}(length(v), v, inv_v)
```
andreasnoack, line 15: I think it's preferable to convert `v` here such that we avoid the a non-concrete type parameter.

devmotion:
An alternative would be to remove the `inv_diag` field from `PDDiagMat` (and similarly `inv_value` from `ScalMat`). From a quick look it seems that it should not be needed if multiplications with `inv_diag` are replaced with divisions by `diag`. This would also avoid the issue in this PR and unwanted promotions: Eg, the fix suggested by @andreasnoack would cause integer-valued diagonal matrices to be promoted to floating point numbers which would affect eg multiplication which should not depend on the type of the inverse (IMO).

src/pdiagmat.jl
Wynand: What would you think of adding an extra type parameter for the inverse? (Like in the code I just pushed)

I'm hesitant to modify type signatures but that would avoid converting ranges to vectors

devmotion: Changing the type parameters makes this PR breaking. I still think we could just remove `inv_diag`. This would also allow us to keep the type signature, and hence the PR would be non-breaking (IMO fields are not part of the public API).

Wynand: Alright, I've pushed commits that naively remove `inv_diag` at the cost of recomputing it whenever it's currently used, or switching to use `diag` instead

To make sure `PDiagMat` is backwards compatible I've also added a custom accessor and constructor, and if these changes make sense I'll update the tests too

devmotion: Oh, I realized that you made this changes only now after I had opened https://github.com/JuliaStats/PDMats.jl/pull/146 (which also changes `ScalMat`). This PR never computes `map(inv, diag)`, so there are no additional allocations (in fact, it reduces allocations since the inverse is not computed anymore when `PDiagMat` is constructed).

Wynand: I think your MR is a better change, so when it merges we can close this one

devmotion:
Fixed by https://github.com/JuliaStats/PDMats.jl/pull/146.

PR #142, opened 2021-11-03
introduce `@testset`s

st--:
@devmotion here's the testsets :)


andreasnoack (approved):
Thanks for adding these.

devmotion:
Thanks @st--!

PR #144, opened 2021-11-03
Avoid copy in `chol_lower`

### LINKED ISSUE #143: `chol_lower` does not avoid allocations
https://github.com/JuliaStats/PDMats.jl/blob/dfcf96c90ade0ad6052855a3e4b3488b3dbe002a/src/chol.jl#L2 always has a superfluous copy, even though the point of defining `chol_lower` in this package is to avoid the extra allocation (as mentioned in the comment a few lines further down), to work around the issue that https://github.com/JuliaLang/julia/pull/42920 tries to address upstream.

MWE:
```julia
julia> using LinearAlgebra, PDMats

julia> A = rand(100, 100); C = A'A;

julia> PDMats.chol_lower(C); PDMats.chol_lower(cholesky(C)); cholesky(C).U';

julia> @allocated PDMats.chol_lower(C)
160240

julia> @allocated PDMats.chol_lower(cholesky(C))
80192

julia> @allocated cholesky(C).U'
80192
```

st--:
Fixes #143

### DIFF
--- a/src/chol.jl
+++ b/src/chol.jl
@@ -1,9 +1,10 @@
 CholType{T,S<:AbstractMatrix} = Cholesky{T,S}
-chol_lower(a::Matrix) = cholesky(a).L
 
 # always use the stored cholesky factor, not a copy
 chol_lower(a::CholType) = a.uplo === 'L' ? a.L : a.U'
 chol_upper(a::CholType) = a.uplo === 'U' ? a.U : a.L'
+
+chol_lower(a::Matrix) = chol_lower(cholesky(a))
 
 if HAVE_CHOLMOD
     CholTypeSparse{T} = SuiteSparse.CHOLMOD.Factor{T}
--- a/src/chol.jl
+++ b/src/chol.jl
@@ -1,7 +1,9 @@
+CholType{T,S<:AbstractMatrix} = Cholesky{T,S}
+
 # Accessing a.L directly might involve an extra copy();
 # instead, always use the stored Cholesky factor:
-chol_lower(a::Cholesky) = a.uplo === 'L' ? a.L : a.U'
-chol_upper(a::Cholesky) = a.uplo === 'U' ? a.U : a.L'
+chol_lower(a::CholType) = a.uplo === 'L' ? a.L : a.U'
+chol_upper(a::CholType) = a.uplo === 'U' ? a.U : a.L'
 
 chol_lower(a::Matrix) = chol_lower(cholesky(a))
 
--- a/src/chol.jl
+++ b/src/chol.jl
@@ -5,7 +5,8 @@ CholType{T,S<:AbstractMatrix} = Cholesky{T,S}
 chol_lower(a::CholType) = a.uplo === 'L' ? a.L : a.U'
 chol_upper(a::CholType) = a.uplo === 'U' ? a.U : a.L'
 
-chol_lower(a::Matrix) = chol_lower(cholesky(a))
+# For a dense Matrix, the following allows us to avoid the Adjoint wrapper:
+chol_lower(a::Matrix) = cholesky(Hermitian(a, :L)).L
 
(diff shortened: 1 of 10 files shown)


st--:
Side note: `chol_lower(::CholTypeSparse)` *seems* to be working fine, but I'm not sure I understand enough about the implementation details of SuiteSparse to guarantee this.

src/chol.jl
```diff
@@ -1,9 +1,9 @@
-CholType{T,S<:AbstractMatrix} = Cholesky{T,S}
```
st--, line 1: This seems to be entirely redundant, so I changed it to use Cholesky directly instead.

andreasnoack (commented):
I think you can just make this `chol_lower(a::Matrix) = cholesky(Hermitian(a, :L)).L`.

I agree that `CholType` is redundant but please make that change in a separate PR.

src/chol.jl
st--: Moved into #145

st--:
> I agree that `CholType` is redundant but please make that change in a separate PR.

Done, see #145

st--:
> I think you can just make this `chol_lower(a::Matrix) = cholesky(Hermitian(a, :L)).L`.

Writing it in terms of `chol_lower(cholesky(a))` seems cleaner to me, and it should compile to the same code; what benefits do you see in writing it as this ?

andreasnoack:
> and it should compile to the same code

No. The default is upper
```julia
julia> cholesky(A).uplo
'U': ASCII/Unicode U+0055 (category Lu: Letter, uppercase)

julia> cholesky(Hermitian(A, :L)).uplo
'L': ASCII/Unicode U+004C (category Lu: Letter, uppercase)
```

andreasnoack:
It's just an artifact of the not wrapping it in a function
```julia
julia> chol_lower_an(A::Matrix) = cholesky(Hermitian(A, :L)).L;

julia> @allocated chol_lower_an(A);

julia> @allocated chol_lower_an(A)
80144
```
It's also best to avoid wrapper types when possible.

st--:
> It's just an artifact of the not wrapping it in a function

Ah, thanks for pointing that out.

> It's also best to avoid wrapper types when possible.

Could you expand on what you mean ?

andreasnoack:
> Could you expand on what you mean ?

Sure. The more types layers you have the less likely it is that a specific method has been written for it and you risk hitting a less optimized fallback. The compiler can also sometimes completely remove these wrapper types but I believe that the more layers you have, the harder it is for the compiler to do so.

st--:
Ok, so do I understand correctly, you would prefer
```diff
-chol_lower(a::Matrix) = chol_lower(cholesky(a))
+chol_lower(a::Matrix) = cholesky(Hermitian(a, :L)).L
```
because the former returns a `LowerTriangular{Adjoint{Matrix}}`, whereas the latter returns a `LowerTriangular{Matrix}` directly?

andreasnoack:
That is correct

st--:
Thanks, that makes sense. I've changed it accordingly:)

andreasnoack (approved):

st--:
Downside of merging it in with `Hermitian()` is that this breaks AD :S

andreasnoack:
Oh no. Do you have an example?

st--:
Yes, it turns out to be a Zygote issue, I've made a PR to fix it here: https://github.com/FluxML/Zygote.jl/pull/1114

st--:
@andreasnoack looks like the Zygote issue might be a bit more complex to resolve; how about (at least temporarily) changing it from `Hermitian` to `Symmetric`?

andreasnoack:
Given that we have this restriction https://github.com/JuliaStats/PDMats.jl/blob/635a957abc71aa9b63d276f96a185db1bed9de74/src/PDMats.jl#L36 it should be fine

PR #146, opened 2021-11-04
Remove `inv_value` and `inv_diag`

devmotion:
Alternative to https://github.com/JuliaStats/PDMats.jl/pull/141 that removes `inv_value` and `inv_diag` fields from `ScalMat` and `PDiagMat`.

Advantages compared with the current master branch are that there are no problems (incompatible types or undesired type promotions) with vectors and values where the inverse are of different type. Two helper functions are added that use division instead of operating with `inv(value)` directly to avoid undesired type promotions (e.g., if `x isa AbstractVector{Float32}` and `value isa Int`, then `inv(value) * x isa AbstractVector{Float64}` whereas `x / value isa AbstractVector{Float32}`).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "PDMats"
 uuid = "90014a1f-27ba-587c-ab20-58faa44d9150"
-version = "0.11.3"
+version = "0.11.4"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/pdiagmat.jl
+++ b/src/pdiagmat.jl
@@ -1,21 +1,12 @@
 """
 Positive definite diagonal matrix.
 """
-struct PDiagMat{T<:Real,V<:AbstractVector} <: AbstractPDMat{T}
+struct PDiagMat{T<:Real,V<:AbstractVector{T}} <: AbstractPDMat{T}
     dim::Int
     diag::V
-    inv_diag::V
-
-    PDiagMat{T,S}(d::Int,v::AbstractVector,inv_v::AbstractVector) where {T,S} =
-        new{T,S}(d,v,inv_v)
 end
 
-function PDiagMat(v::AbstractVector,inv_v::AbstractVector)
-    @check_argdims length(v) == length(inv_v)
-    PDiagMat{eltype(v),typeof(v)}(length(v), v, inv_v)
-end
-
-PDiagMat(v::AbstractVector) = PDiagMat(v, inv.(v))
+PDiagMat(v::AbstractVector{<:Real}) = PDiagMat{eltype(v),typeof(v)}(length(v), v)
 
 ### Conversion
 Base.convert(::Type{PDiagMat{T}},      a::PDiagMat) where {T<:Real} = PDiagMat(convert(AbstractArray{T}, a.diag))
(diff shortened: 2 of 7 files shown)


src/pdiagmat.jl
```diff
@@ -132,7 +127,7 @@ end
 ### tri products
 
 function X_A_Xt(a::PDiagMat, x::StridedMatrix)
-    z = x .* reshape(sqrt.(a.diag), 1, dim(a))
+    z = x .* sqrt.(reshape(a.diag, 1, dim(a)))
```
devmotion, line 130: This (and similar line below) should be fixed in any case. It seems the current implementation prevents broadcast fusion and creates an additional intermediate array.

src/scalmat.jl
```diff
@@ -92,10 +89,10 @@ end
 
 function X_invA_Xt(a::ScalMat, x::StridedMatrix)
     @check_argdims dim(a) == size(x, 2)
-    lmul!(a.inv_value, x * transpose(x))
+    ldiv!(a.value, x * transpose(x))
```
devmotion, line 92: It seems `ldiv!(::Number, ::AbstractArray)` is not available on Julia 1.0. I did not realize this since there is no note in the docstring: https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.ldiv!

devmotion:
@Wynand I also added your failing example as a test.

andreasnoack (approved):

andreasnoack:
Should we consider this a breaking change?

andreasnoack:
Should we consider this a breaking change?

devmotion:
It is supposed to be non-breaking - I deprecated the constructors but considered the fields to be not part of the public API (alternatively, I could add a deprecation warning in `getproperty`). But maybe it's safer to tag a breaking release?

andreasnoack:
It would at least break https://github.com/devmotion/CalibrationErrorsDistributions.jl/blob/41be6f90f9cf57e3f9e8333d3d1f4eb73d55002e/src/mvnormal.jl#L106 which might be familiar to you so maybe let's bump the minor version.

andreasnoack:
I went through all the direct dependencies and yours is the only one that touches the field so I'd also be okay with just making it a patch release.

devmotion:
Ah I'd like to avoid breaking my package, even if it accesses (internal?) fields :smile: I'll update the package in a way that does not use `inv_diag`/`inv_value` first before merging this PR.

devmotion:
I fixed CalibrationErrorsDistributions and updated the README which I forgot initially.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "PDMats"
 uuid = "90014a1f-27ba-587c-ab20-58faa44d9150"
-version = "0.11.3"
+version = "0.11.4"
```
Wynand, line 3: Should this have bumped to 0.11.5?  It looks like 0.11.4 was already released when this was merged: https://github.com/JuliaStats/PDMats.jl/releases/tag/v0.11.4

devmotion: Yes, it should, JuliaRegistrator noticed it as well: https://github.com/JuliaStats/PDMats.jl/commit/2f3a83e4f27fae1ba9592f330a9cbaec4bb60284#commitcomment-60431201 The PR was created before the other PR was merged and released, so the diff in the PR was outdated and based on an earlier version of the master branch.

devmotion: I fixed it: https://github.com/JuliaStats/PDMats.jl/commit/15fecad14e700e024d359733d101c992707e981b :slightly_smiling_face:

PR #152, opened 2022-01-27
Improve StaticArray + BandedMatrices support (update of #131)

### LINKED ISSUE #151: Support for StaticArrays
Are there plans to support `StaticArrays` with `PDMats`?

I'm measuring a 4x speedup on `cholesky()` with a 10x10 `SMatrix` vs `Matrix`, so it seems like this would be worthwhile.

### LINKED ISSUE #154: Support for BandedMatrices
At the moment, `PDMat` casts `BandedMatrix` to a dense array. It would be nice it they could be handled with preserving the structure.

devmotion:
I updated #131, resolved the merge conflicts, and extended the tests which revealed some other methods that had to be fixed since currently they don't support StaticArrays.

~~I also fixed https://github.com/JuliaStats/PDMats.jl/pull/131#issuecomment-805893449, made it consistent with the behaviour in LinearAlgebra (matrix for vectors of 1 element), added tests, and in particular added checks of the dimensions of the input such that we don't return incorrect results for dimensions > 1 anymore.~~

~~Possibly the PR should be split but initially it was a bit unclear to me what should be removed since all changes are motivated by and fix errors in the StaticArray test set.~~

cc @chriselrod @andreasnoack 

Fixes #151. Fixes #154.

Edit: Split PR and moved the non-StaticArrays fixes to https://github.com/JuliaStats/PDMats.jl/pull/153.


devmotion:
Test revealed additional bugs: for CHOLMOD `/(::Matrix, ::CholTypeSparse)` is not defined, and hence `/(::Matrix, ::PDSparseMat)` errors, and on older Julia versions (eg 1.0) also `/(::Matrix, ::Cholesky)` is not defined and hence `/(::Matrix, ::PDMat)` errors. Both bugs have existed for some time but there were no tests for `/`.

The `Cholesky`/`PDMat` issue could be fixed by adding an internal function which uses the method in base if it exists and otherwise uses our copy of it (similar to what we do for `rdiv!`). Alternatively, we could not drop support for older Julia versions that don't support.

I'm not sure what do with CHOLMOD. For now I just skipped the relevant tests.

devmotion:
OK, this PR is much simpler now 🙂 It contains only a minimal set of changes and should be ready for review @andreasnoack

nalimilan:
> I also fixed [#131 (comment)](https://github.com/JuliaStats/PDMats.jl/pull/131#issuecomment-805893449), made it consistent with the behaviour in LinearAlgebra (matrix for vectors of 1 element), added tests, and in particular added checks of the dimensions of the input such that we don't return incorrect results for dimensions > 1 anymore.

This seems to be potentially breaking, but I don't see it in the diff. Have you dropped this part?

devmotion:
I moved it to a separate PR, it's already merged and released (as a non-breaking release since we considered it a bugfix).

devmotion:
I update the initial comment.

nalimilan (approved):
I'm not really qualified but this looks really simple and well tested, so...

PR #153, opened 2022-02-07
Fix right division and add checks of dimensions

devmotion:
To simplify reviewing and to avoid unrelated changes, I extracted the fixes of right division and additional checks of dimensions from https://github.com/JuliaStats/PDMats.jl/pull/152.

This addresses the issues discussed in https://github.com/JuliaStats/PDMats.jl/pull/131#issuecomment-805893449 and makes the behaviour of `/` consistent with LinearAlgebra. Moreover, the dimension checks prevent incorrect behaviour, e.g., currently the following "works":
```julia
julia> using PDMats

julia> ScalMat(4, 2.0) * ones(3)
3-element Vector{Float64}:
 2.0
 2.0
 2.0

julia> ScalMat(4, 2.0) * ones(3, 2)
3×2 Matrix{Float64}:
 2.0  2.0
 2.0  2.0
 2.0  2.0
```

### DIFF
--- a/test/pdmtypes.jl
+++ b/test/pdmtypes.jl
@@ -83,4 +83,30 @@ using Test
         @test d isa PDiagMat{eltype(v),typeof(v)}
         @test d.diag === v
     end
+
+    @testset "division of vectors (dim = 1)" begin
+        A = rand(1, 1)
+        x = randn(1)
+        y = x / A
+        @assert x / A isa Matrix{Float64}
+        @assert size(y) == (1, 1)
+
+        for M in (PDiagMat(vec(A)), ScalMat(1, first(A)))
+            @test x / M isa Matrix{Float64}
+            @test x / M ≈ y
+        end
+
+        # requires https://github.com/JuliaLang/julia/pull/32594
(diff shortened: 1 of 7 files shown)


devmotion:
Bump :slightly_smiling_face:

andreasnoack (approved):

test/pdmtypes.jl
```diff
@@ -83,4 +83,30 @@ using Test
         @test d isa PDiagMat{eltype(v),typeof(v)}
         @test d.diag === v
     end
+
+    @testset "division of vectors (dim = 1)" begin
+        A = rand(1, 1)
+        x = randn(1)
+        y = x / A
+        @assert x / A isa Matrix{Float64}
```
KristofferC, line 91: @devmotion, This is no longer true on 1.9:

```
julia> y = x / A
1-element Vector{Float64}:
 1.1240182465158486
```

devmotion: Do you remember which PR/commit changed this behaviour? I quickly looked through the Julia repo but did not immediately find it.

Issue #136, opened 2021-08-12
Unifying constructor? e.g. PDMat(Diagonal(...)) = PDiagMat(diag(...))


st--:
Would be really helpful in downstream code if I wouldn't have to think whether I've got a `Diagonal` -> wrap in `PDiagMat` or not -> wrap in `PDMat`. How about
```julia
PDMat(mat::Diagonal) = PDiagMat(diag(mat))
```
for example?

PR #156, opened 2022-03-14
Add support for Symmetric{<:Any,<:Diagonal}

chriselrod:
On master, this throws:
```julia
  MethodError: Cannot `convert` an object of type Symmetric{Float64, Diagonal{Float64, Vector{Float64}}} to an object of type Diagonal{Float64, Vector{Float64}}
```

### DIFF
--- a/src/pdmat.jl
+++ b/src/pdmat.jl
@@ -17,6 +17,7 @@ function PDMat(mat::AbstractMatrix,chol::Cholesky{T,S}) where {T,S}
 end
 
 PDMat(mat::AbstractMatrix) = PDMat(mat, cholesky(mat))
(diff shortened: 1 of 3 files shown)


chriselrod:
Thoughts -- should I add methods to support older Julia versions, or just place support for this behind Julia version checks (i.e. >= 1.8)?

devmotion:
Related: https://github.com/JuliaStats/PDMats.jl/issues/136

chriselrod:
Ah, I guess the issue is in Distributions.jl then.

devmotion:
Yes, IMO this is weird since `PDMat(...)` seems to imply that you create an instance of type `PDMat`. Maybe one could use `AbstractPDMat` or `pdmat`...

PR #158, opened 2022-04-28
Implement matrix square root for PDMats

wil-j-wil:
The matrix square root, [`LinearAlgebra.sqrt()`](https://github.com/JuliaLang/julia/blob/dcc0efe26fd0490bf963be8d50dc6df83b91a868/stdlib/LinearAlgebra/src/dense.jl#L806), is not currently implemented for `PDMats`, i.e., `sqrt(::AbstractPDMat)` throws an error.

This PR implements this by applying `sqrt(Matrix(A))` in the generic case, `sqrt(A.mat)` where possible, plus specific cases for `PDiagMat`, `ScaleMat` and `PDSparseMat`.

Tests also written which check the specific cases return the same result as `sqrt(Matrix(A))`, which I believe is sufficient.

### DIFF
--- a/src/generics.jl
+++ b/src/generics.jl
@@ -28,6 +28,13 @@ Base.kron(A::AbstractPDMat, B::AbstractPDMat) = PDMat(kron(Matrix(A), Matrix(B))
 LinearAlgebra.isposdef(::AbstractPDMat) = true
 LinearAlgebra.ishermitian(::AbstractPDMat) = true
 
+"""
+    sqrt(A::AbstractPDMat)
+
+Return the matrix square root of `A`: the matrix `X` such that ``XX = A``
+"""
+LinearAlgebra.sqrt(A::AbstractPDMat) = PDMat(sqrt(Matrix(A)))
+
 ## whiten and unwhiten
 whiten!(a::AbstractPDMat, x::StridedVecOrMat) = whiten!(x, a, x)
 unwhiten!(a::AbstractPDMat, x::StridedVecOrMat) = unwhiten!(x, a, x)
--- a/src/generics.jl
+++ b/src/generics.jl
@@ -28,11 +28,7 @@ Base.kron(A::AbstractPDMat, B::AbstractPDMat) = PDMat(kron(Matrix(A), Matrix(B))
 LinearAlgebra.isposdef(::AbstractPDMat) = true
 LinearAlgebra.ishermitian(::AbstractPDMat) = true
 
-"""
-    sqrt(A::AbstractPDMat)
-
-Return the matrix square root of `A`: the matrix `X` such that ``XX = A``
-"""
+# Generic fallback
 LinearAlgebra.sqrt(A::AbstractPDMat) = PDMat(sqrt(Matrix(A)))
 
 ## whiten and unwhiten
--- a/src/generics.jl
+++ b/src/generics.jl
@@ -28,9 +28,6 @@ Base.kron(A::AbstractPDMat, B::AbstractPDMat) = PDMat(kron(Matrix(A), Matrix(B))
 LinearAlgebra.isposdef(::AbstractPDMat) = true
 LinearAlgebra.ishermitian(::AbstractPDMat) = true
 
-# Generic fallback
-LinearAlgebra.sqrt(A::AbstractPDMat) = PDMat(sqrt(Matrix(A)))
-
 ## whiten and unwhiten
 whiten!(a::AbstractPDMat, x::StridedVecOrMat) = whiten!(x, a, x)
 unwhiten!(a::AbstractPDMat, x::StridedVecOrMat) = unwhiten!(x, a, x)
--- a/src/pdmat.jl
+++ b/src/pdmat.jl
@@ -57,7 +57,7 @@ LinearAlgebra.logdet(a::PDMat) = logdet(a.chol)
 LinearAlgebra.eigmax(a::PDMat) = eigmax(a.mat)
 LinearAlgebra.eigmin(a::PDMat) = eigmin(a.mat)
 Base.kron(A::PDMat, B::PDMat) = PDMat(kron(A.mat, B.mat), Cholesky(kron(A.chol.U, B.chol.U), 'U', A.chol.info))
-LinearAlgebra.sqrt(A::PDMat) = PDMat(sqrt(A.mat))
+LinearAlgebra.sqrt(A::PDMat) = PDMat(sqrt(Hermitian(A.mat)))
 
 ### whiten and unwhiten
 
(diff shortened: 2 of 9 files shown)


src/generics.jl
```diff
@@ -28,6 +28,13 @@ Base.kron(A::AbstractPDMat, B::AbstractPDMat) = PDMat(kron(Matrix(A), Matrix(B))
 LinearAlgebra.isposdef(::AbstractPDMat) = true
 LinearAlgebra.ishermitian(::AbstractPDMat) = true
 
+"""
+    sqrt(A::AbstractPDMat)
+
+Return the matrix square root of `A`: the matrix `X` such that ``XX = A``
+"""
```
devmotion, line 35: This should not be documented according to the Julia documentation since it does not differ from the documentation for `sqrt(::AbstractMatrix)` (https://github.com/JuliaLang/julia/blob/dcc0efe26fd0490bf963be8d50dc6df83b91a868/stdlib/LinearAlgebra/src/dense.jl#L806) and `AbstractPDMat <: AbstractMatrix`.

```suggestion
# Generic fallback
```

src/generics.jl
```diff
@@ -28,6 +28,13 @@ Base.kron(A::AbstractPDMat, B::AbstractPDMat) = PDMat(kron(Matrix(A), Matrix(B))
...
+LinearAlgebra.sqrt(A::AbstractPDMat) = PDMat(sqrt(Matrix(A)))
```
devmotion, line 36: I'm a bit uneasy with this fallback. For special input types usually you don't want a `PDMat` as result, as already PDiagMat and ScalMat show. What's even worse apart from possible inefficiences is that sometimes `PDMat` will even error (e.g., if you use positive-semidefinite matrices as supported and defined in `PDMatsExtras`).

So I think, either we should just use
```suggestion
LinearAlgebra.sqrt(A::AbstractPDMat) = sqrt(Hermitian(Matrix(A)))
```
(by wrapping the matrix with `Hermitian` the eigen decomposition is used: https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#Base.sqrt-Tuple{StridedMatrix{T}%20where%20T}) or just not define a fallback at all.

Personally, I tend towards the second option.

src/pdmat.jl
```diff
@@ -57,6 +57,7 @@ LinearAlgebra.logdet(a::PDMat) = logdet(a.chol)
 LinearAlgebra.eigmax(a::PDMat) = eigmax(a.mat)
 LinearAlgebra.eigmin(a::PDMat) = eigmin(a.mat)
 Base.kron(A::PDMat, B::PDMat) = PDMat(kron(A.mat, B.mat), Cholesky(kron(A.chol.U, B.chol.U), 'U', A.chol.info))
+LinearAlgebra.sqrt(A::PDMat) = PDMat(sqrt(A.mat))
```
devmotion, line 60:
```suggestion
LinearAlgebra.sqrt(A::PDMat) = PDMat(sqrt(Hermitian(A.mat)))
```

src/pdsparsemat.jl
```diff
@@ -52,6 +52,7 @@ end
 
 Base.inv(a::PDSparseMat{T}) where {T<:Real} = PDMat(inv(a.mat))
 LinearAlgebra.logdet(a::PDSparseMat) = logdet(a.chol)
+LinearAlgebra.sqrt(A::PDSparseMat) = PDSparseMat(sparse(sqrt(Matrix(A))))
```
devmotion, line 55: I would have assumed that the square root of a sparse matrix is not necessarily sparse? So if we want to wrap it with a `AbstractPDMat` maybe we should rather define
```suggestion
LinearAlgebra.sqrt(A::PDSparseMat) = PDMat(sqrt(Hermitian(Matrix(A))))
```

test/sqrt.jl
```diff
@@ -0,0 +1,44 @@
+using PDMats
+using Test
+using LinearAlgebra: LinearAlgebra
+
+_randPDMat(T, n) = (X = randn(T, n, n); PDMat(X * X' + LinearAlgebra.I))
+_randPDiagMat(T, n) = PDiagMat(rand(T, n))
+_randScalMat(T, n) = ScalMat(n, rand(T))
+_randPDSparseMat(T, n) = (X = T.(sprand(n, 1, 0.5)); PDSparseMat(X * X' + LinearAlgebra.I))
```
devmotion, line 8: Most of these are already define in https://github.com/JuliaStats/PDMats.jl/blob/84df00b41da9573c3e0b7e79bdcde2dda24bb2d5/test/kron.jl#L5-L7. Can you avoid the method redefinition and move these methods to tests/testutils.jl?

test/sqrt.jl
```diff
@@ -0,0 +1,44 @@
...
+    # specific method required for testing cholesky of a sparse matrix
+    Asqrt = sqrt(A)
+    Asqrt_dense = PDMat(sqrt(Matrix(A)))
+    @test dim(Asqrt) == dim(Asqrt_dense)
+    @test Matrix(Asqrt) ≈ Matrix(Asqrt_dense)
+    cAsqrt = cholesky(Asqrt)
+    perm = cAsqrt.p  # account for permutation
+    L = sparse(cAsqrt.L)
+    @test Asqrt[perm, perm] ≈ L * L'
+end
```
devmotion, line 33: Is it possible to use https://github.com/JuliaStats/PDMats.jl/blob/84df00b41da9573c3e0b7e79bdcde2dda24bb2d5/test/testutils.jl#L95-L121 etc. here?

devmotion (commented):
Thanks for the PR! I think it's a good idea and we should add definitions for `sqrt`. I left some comments.

src/generics.jl
wil-j-wil: Yep, I'm happy to remove the fallback.

test/sqrt.jl
wil-j-wil: Good spot. Have moved to `testutils.jl`

test/sqrt.jl
```diff
@@ -0,0 +1,26 @@
+using PDMats
+using Test
+using LinearAlgebra: LinearAlgebra
+
+_randPDSparseMat(T, n) = (X = T.(sprand(n, 1, 0.5)); PDSparseMat(X * X' + LinearAlgebra.I))
```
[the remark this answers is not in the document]
wil-j-wil: done

test/testutils.jl
```diff
@@ -319,3 +319,17 @@ function pdtest_whiten(C::AbstractPDMat, Cmat::Matrix, verbose::Int)
...
+_randPDMat(T, n) = (X = randn(T, n, n); PDMat(X * X' + LinearAlgebra.I))
+_randPDiagMat(T, n) = PDiagMat(rand(T, n))
+_randScalMat(T, n) = ScalMat(n, rand(T))
+
+function _pd_compare(A::AbstractPDMat, B::AbstractPDMat)
+    @test dim(A) == dim(B)
+    @test Matrix(A) ≈ Matrix(B)
+    @test cholesky(A).L ≈ cholesky(B).L
+    @test cholesky(A).U ≈ cholesky(B).U
+end
```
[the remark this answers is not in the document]
wil-j-wil: True. I have removed `_pd_compare` from the sqrt tests completely, since `pdtest_cmat()` is sufficient.

test/sqrt.jl
wil-j-wil: the tests now rely entirely on these

devmotion (approved):
Looks good to me, thanks!

PR #162, opened 2022-05-19
Implement fallbacks for (un)whiten(!) and (inv)quad(!) to AbstractArray

oxinabox:
This is the partner to https://github.com/JuliaStats/Distributions.jl/pull/1552#issuecomment-1130483202
which will allow resolving https://github.com/JuliaStats/Distributions.jl/issues/1219

It doesn't go and implement everything falling back to AbstractArray,
just the operations required for Multivariate Normals,
and their family (so I can test them)

In some cases that implementation is just making the PDMat one generic, (since `cholesky(x::PDMat)=x.chol` is defined, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ While in theory all of them can be defined, at present only the following subset
  - `quad`, `quad!`
  - `invquad`, `invquad!`
 
-PRs to implement more generic fallbacks are welcome
+PRs to implement more generic fallbacks are welcome.
 
 ## Define Customized Subtypes
 
--- a/README.md
+++ b/README.md
@@ -209,6 +209,18 @@ unwhiten!(a, x)     # un-whitening transform inplace, updating `x`.
 unwhiten!(r, a, x)  # write the transformed result to `r`.
 ```
 
+### Fallbacks for `AbstractArray`s
+For ease of composability, some of these functions have generic fallbacks defined that work on `AbstractArray`s.
+These fallbacks may not be as fast as the methods specializaed for `AbstractPDMat`s, but they let you more easily swap out types.
+While in theory all of them can be defined, at present only the following subset has:
+
+ - `dim`
+ - `whiten`, `whiten!`
+ - `unwhiten`, `unwhiten!`
+ - `quad`, `quad!`
+ - `invquad`, `invquad!`
+
+PRs to implement more generic fallbacks are welcome
 
 ## Define Customized Subtypes
 
--- a/src/generics.jl
+++ b/src/generics.jl
@@ -48,15 +48,10 @@ function unwhiten!(r::AbstractVecOrMat, a::AbstractMatrix, x::AbstractVecOrMat)
 end
 
 """
-    whiten(a::AbstractMatrix, x::AbstractVecOrMat)    
-    whiten(a::AbstractPDMat, x::StridedVecOrMat)
-    whiten!(a::AbstractPDMat, x::StridedVecOrMat)
-    whiten!(r::StridedVecOrMat, a::AbstractPDMat, x::StridedVecOrMat)
+    whiten(a::AbstractMatrix, x::AbstractVecOrMat)
     unwhiten(a::AbstractMatrix, x::AbstractVecOrMat)
-    unwhiten(a::AbstractPDMat, x::StridedVecOrMat)
     unwhiten!(a::AbstractMatrix, x::AbstractVecOrMat)
-    unwhiten!(a::AbstractPDMat, x::StridedVecOrMat)
-    unwhiten!(r::StridedVecOrMat, a::AbstractPDMat, x::StridedVecOrMat)
+    unwhiten!(r::AbstractVecOrMat, a::AbstractPDMat, x::AbstractVecOrMat)
 
 Allocating and in-place versions of the `whiten`ing transform (or its inverse) defined by `a` applied to `x`
 
--- a/src/generics.jl
+++ b/src/generics.jl
@@ -114,7 +114,6 @@ invquad(a::AbstractMatrix, x::AbstractVector) = sum(abs2, chol_lower(cholesky(a)
 Overwrite `r` with the value of the quadratic form defined by `a` applied columnwise to `x`
 """
 quad!(r::AbstractArray, a::AbstractMatrix, x::AbstractMatrix) = colwise_dot!(r, x, a * x)
-invquad!(r::AbstractArray, a::AbstractMatrix, x::AbstractMatrix) = colwise_dot!(r, x, a \ x)
 
 
 """
--- a/src/generics.jl
+++ b/src/generics.jl
@@ -94,7 +94,6 @@ unwhiten(a::AbstractMatrix, x::AbstractVecOrMat) = unwhiten!(similar(x), a, x)
 
 """
     quad(a::AbstractMatrix, x::AbstractVecOrMat)
-    quad(a::AbstractPDMat, x::StridedVecOrMat)
 
 Return the value of the quadratic form defined by `a` applied to `x`
 
@@ -110,8 +109,7 @@ quad(a::AbstractMatrix, x::AbstractVector) = sum(abs2, chol_upper(cholesky(a)) *
 invquad(a::AbstractMatrix, x::AbstractVector) = sum(abs2, chol_lower(cholesky(a)) \ x)
 
 """
-    quad!(r::AbstractArray, a::AbstractMatrix, x::AbstractMatrix
-    quad!(r::AbstractArray, a::AbstractPDMat, x::StridedMatrix)
+    quad!(r::AbstractArray, a::AbstractMatrix, x::AbstractMatrix)
 
 Overwrite `r` with the value of the quadratic form defined by `a` applied columnwise to `x`
 """
@@ -121,7 +119,6 @@ invquad!(r::AbstractArray, a::AbstractMatrix, x::AbstractMatrix) = colwise_dot!(
 
 """
     invquad(a::AbstractMatrix, x::AbstractVecOrMat)
-    invquad(a::AbstractPDMat, x::StridedVecOrMat)
 
 Return the value of the quadratic form defined by `inv(a)` applied to `x`.
 
@@ -138,7 +135,6 @@ end
 
 """
     invquad!(r::AbstractArray, a::AbstractMatrix, x::AbstractMatrix)
-    invquad!(r::AbstractArray, a::AbstractPDMat, x::StridedMatrix)
 
 Overwrite `r` with the value of the quadratic form defined by `inv(a)` applied columnwise to `x`
 """
--- a/src/generics.jl
+++ b/src/generics.jl
@@ -1,10 +1,14 @@
 # Generic functions (on top of the type-specific implementations)
 
 ## Basic functions
-
 Base.size(a::AbstractPDMat) = (dim(a), dim(a))
 Base.size(a::AbstractPDMat, i::Integer) = 1 <= i <= 2 ? dim(a) : 1
 Base.length(a::AbstractPDMat) = abs2(dim(a))
+
+function dim(a::AbstractMatrix)
+    @check_argdims size(a, 1) == size(a, 2)
+    return size(a, 1)
+end
 
 ## arithmetics
 
@@ -29,14 +33,28 @@ LinearAlgebra.isposdef(::AbstractPDMat) = true
 LinearAlgebra.ishermitian(::AbstractPDMat) = true
 
 ## whiten and unwhiten
-whiten!(a::AbstractPDMat, x::StridedVecOrMat) = whiten!(x, a, x)
-unwhiten!(a::AbstractPDMat, x::StridedVecOrMat) = unwhiten!(x, a, x)
+
+whiten!(a::AbstractMatrix, x::AbstractVecOrMat) = whiten!(x, a, x)
+unwhiten!(a::AbstractMatrix, x::AbstractVecOrMat) = unwhiten!(x, a, x)
+
+function whiten!(r::AbstractVecOrMat, a::AbstractMatrix, x::AbstractVecOrMat)
+    v = _rcopy!(r, x)
+    ldiv!(chol_lower(cholesky(a)), v)
+end
+
+function unwhiten!(r::AbstractVecOrMat, a::AbstractMatrix, x::AbstractVecOrMat)
+    v = _rcopy!(r, x)
+    lmul!(chol_lower(cholesky(a)), v)
+end
 
 """
+    whiten(a::AbstractMatrix, x::AbstractVecOrMat)    
     whiten(a::AbstractPDMat, x::StridedVecOrMat)
     whiten!(a::AbstractPDMat, x::StridedVecOrMat)
     whiten!(r::StridedVecOrMat, a::AbstractPDMat, x::StridedVecOrMat)
+    unwhiten(a::AbstractMatrix, x::AbstractVecOrMat)
     unwhiten(a::AbstractPDMat, x::StridedVecOrMat)
+    unwhiten!(a::AbstractMatrix, x::AbstractVecOrMat)
     unwhiten!(a::AbstractPDMat, x::StridedVecOrMat)
     unwhiten!(r::StridedVecOrMat, a::AbstractPDMat, x::StridedVecOrMat)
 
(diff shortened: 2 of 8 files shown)


src/pdmat.jl
```diff
@@ -60,39 +60,6 @@ LinearAlgebra.eigmin(a::PDMat) = eigmin(a.mat)
 Base.kron(A::PDMat, B::PDMat) = PDMat(kron(A.mat, B.mat), Cholesky(kron(A.chol.U, B.chol.U), 'U', A.chol.info))
 LinearAlgebra.sqrt(A::PDMat) = PDMat(sqrt(Hermitian(A.mat)))
 
-### whiten and unwhiten
```
oxinabox, line 63: The below have been made generic and moved to generics.jl

src/utils.jl
```diff
@@ -69,11 +69,10 @@ function invwsumsq(w::AbstractVector, a::AbstractVector)
...
 
 function colwise_dot!(r::AbstractArray, a::AbstractMatrix, b::AbstractMatrix)
-    n = length(r)
-    @check_argdims n == size(a, 2) == size(b, 2) && size(a, 1) == size(b, 1)
-    for j = 1:n
+    @check_argdims size(a, 2) == size(b, 2) && size(a, 1) == size(b, 1)
+    for j in axes(a, 2)
         v = zero(promote_type(eltype(a), eltype(b)))
-        @simd for i = 1:size(a, 1)
+        @simd for i in axes(a, 1)
```
oxinabox, line 75: Important modernizing julia code so it works with OffsetArrays etc

src/generics.jl
```diff
@@ -68,27 +86,41 @@ julia> W * W'
  0.0  1.0
 ```
 """
-whiten(a::AbstractPDMat, x::StridedVecOrMat) = whiten!(similar(x), a, x)
-unwhiten(a::AbstractPDMat, x::StridedVecOrMat) = unwhiten!(similar(x), a, x)
+whiten(a::AbstractMatrix, x::AbstractVecOrMat) = whiten!(similar(x), a, x)
```
devmotion, line 89: Unrelated, but would probably be good to not fall back on the in-place methods here to avoid e.g. creating `MArray` from `SArray`s (IIRC we already did this for `PDMat` etc.).

test/runtests.jl
```diff
@@ -1,5 +1,5 @@
 include("testutils.jl")
-tests = ["pdmtypes", "addition", "generics", "kron", "chol", "specialarrays", "sqrt"]
+tests = ["pdmtypes", "abstracttypes.jl", "addition", "generics", "kron", "chol", "specialarrays", "sqrt"]
```
devmotion, line 2:
```suggestion
tests = ["pdmtypes", "abstracttypes", "addition", "generics", "kron", "chol", "specialarrays", "sqrt"]
```

src/utils.jl
devmotion: Yes, probably cleaner in a separate PR.

README.md
```diff
@@ -209,6 +209,18 @@ unwhiten!(a, x)     # un-whitening transform inplace, updating `x`.
...
+These fallbacks may not be as fast as the methods specializaed for `AbstractPDMat`s, but they let you more easily swap out types.
+While in theory all of them can be defined, at present only the following subset has:
+
+ - `dim`
+ - `whiten`, `whiten!`
+ - `unwhiten`, `unwhiten!`
+ - `quad`, `quad!`
+ - `invquad`, `invquad!`
+
+PRs to implement more generic fallbacks are welcome
```
devmotion, line 223:
```suggestion
PRs to implement more generic fallbacks are welcome.
```

src/generics.jl
```diff
@@ -29,14 +33,28 @@ LinearAlgebra.isposdef(::AbstractPDMat) = true
...
+end
+
+function unwhiten!(r::AbstractVecOrMat, a::AbstractMatrix, x::AbstractVecOrMat)
+    v = _rcopy!(r, x)
+    lmul!(chol_lower(cholesky(a)), v)
+end
 
 """
+    whiten(a::AbstractMatrix, x::AbstractVecOrMat)    
     whiten(a::AbstractPDMat, x::StridedVecOrMat)
```
devmotion, line 52: `AbstractPDMat <: AbstractMatrix`, so maybe these signatures can be removed?

src/generics.jl
```diff
@@ -68,27 +86,41 @@ julia> W * W'
...
-unwhiten(a::AbstractPDMat, x::StridedVecOrMat) = unwhiten!(similar(x), a, x)
+whiten(a::AbstractMatrix, x::AbstractVecOrMat) = whiten!(similar(x), a, x)
+unwhiten(a::AbstractMatrix, x::AbstractVecOrMat) = unwhiten!(similar(x), a, x)
 
 
 ## quad
 
 """
+    quad(a::AbstractMatrix, x::AbstractVecOrMat)
     quad(a::AbstractPDMat, x::StridedVecOrMat)
```
devmotion, line 97: Same here, maybe remove this?
```suggestion
```

src/generics.jl
```diff
@@ -68,27 +86,41 @@ julia> W * W'
...
+function quad(a::AbstractMatrix{T}, x::AbstractMatrix{S}) where {T<:Real, S<:Real}
     @check_argdims dim(a) == size(x, 1)
     quad!(Array{promote_type(T, S)}(undef, size(x,2)), a, x)
 end
 
+quad(a::AbstractMatrix, x::AbstractVector) = sum(abs2, chol_upper(cholesky(a)) * x)
+invquad(a::AbstractMatrix, x::AbstractVector) = sum(abs2, chol_lower(cholesky(a)) \ x)
 
 """
+    quad!(r::AbstractArray, a::AbstractMatrix, x::AbstractMatrix
```
devmotion, line 113:
```suggestion
    quad!(r::AbstractArray, a::AbstractMatrix, x::AbstractMatrix)
```

src/generics.jl
```diff
@@ -68,27 +86,41 @@ julia> W * W'
...
+    quad!(r::AbstractArray, a::AbstractPDMat, x::StridedMatrix)
```
devmotion, line 114: Maybe could be removed?
```suggestion
```

src/generics.jl
```diff
@@ -68,27 +86,41 @@ julia> W * W'
...
+
+Overwrite `r` with the value of the quadratic form defined by `a` applied columnwise to `x`
+"""
+quad!(r::AbstractArray, a::AbstractMatrix, x::AbstractMatrix) = colwise_dot!(r, x, a * x)
+invquad!(r::AbstractArray, a::AbstractMatrix, x::AbstractMatrix) = colwise_dot!(r, x, a \ x)
+
+
+"""
+    invquad(a::AbstractMatrix, x::AbstractVecOrMat)
     invquad(a::AbstractPDMat, x::StridedVecOrMat)
```
devmotion, line 124:
```suggestion
```

src/generics.jl
```diff
@@ -98,7 +130,16 @@ For most `PDMat` types this is done in a way that does not require evaluation of
...
-function invquad(a::AbstractPDMat{T}, x::StridedMatrix{S}) where {T<:Real, S<:Real}
+invquad(a::AbstractMatrix, x::AbstractVecOrMat) = x' / a * x
+function invquad(a::AbstractMatrix{T}, x::AbstractMatrix{S}) where {T<:Real, S<:Real}
     @check_argdims dim(a) == size(x, 1)
     invquad!(Array{promote_type(T, S)}(undef, size(x,2)), a, x)
 end
+
+"""
+    invquad!(r::AbstractArray, a::AbstractMatrix, x::AbstractMatrix)
+    invquad!(r::AbstractArray, a::AbstractPDMat, x::StridedMatrix)
```
devmotion, line 141:
```suggestion
```

devmotion (commented):
Looks good to me in principle, but I would like to get someone else's opinion (@andreasnoack maybe?).

When commenting on the function signatures in the docstrings, I noticed that the general definitions for `AbstractMatrix` could cause surprising slow-downs for `AbstractPDMat`s with non-`StridedArray`s if their specializations for `(inv)quad!` etc. are only defined for `StridedArray`s. So maybe it would be good to check if such restrictions still exist and if we can remove them.

oxinabox:
> Looks good to me in principle, but I would like to get someone else's opinion (@andreasnoack maybe?).
> 
> When commenting on the function signatures in the docstrings, I noticed that the general definitions for `AbstractMatrix` could cause surprising slow-downs for `AbstractPDMat`s with non-`StridedArray`s if their specializations for `(inv)quad!` etc. are only defined for `StridedArray`s. So maybe it would be good to check if such restrictions still exist and if we can remove them.

AFAIK those don't slow down. 
They were MethodErrors before and now they are AmbiguityErrors

devmotion:
Are you sure? To me it seems e.g. that https://github.com/JuliaStats/PDMats.jl/blob/92a51785c69a303a6da5142433b7556e1d68d66d/src/pdiagmat.jl#L113 defines an optimization of `quad!` for `a::PDiagMat` and `x::StridedMatrix` but by adding the generic fallback in this PR a call of e.g. `quad!(r::AbstractArray, a::PDiagMat, x::SubArray{Float64,2})` will not use this specialized implementation but the less efficient fallback.

oxinabox:
Looks fine to me:

## With this PR:
```julia
julia> @which invquad!([2,2], PDiagMat([1,1]), [3 3; 3 3])
invquad!(r::AbstractArray, a::PDiagMat, x::StridedMatrix{T} where T) in PDMats at /home/oxinabox/JuliaEnvs/Distributions.jl/dev/PDMats/src/pdiagmat.jl:127

julia> @which invquad!([2,2], PDiagMat([1,1]), @view([3 3; 3 3][:,:]))
invquad!(r::AbstractArray, a::PDiagMat, x::StridedMatrix{T} where T) in PDMats at /home/oxinabox/JuliaEnvs/Distributions.jl/dev/PDMats/src/pdiagmat.jl:127

julia> @which invquad!([2,2], PDiagMat([1,1]), @view([3 3; 3 3][[2 4; 1 3]]))
invquad!(r::AbstractArray, a::AbstractMatrix, x::AbstractMatrix) in PDMats at /home/oxinabox/JuliaEnvs/Distributions.jl/dev/PDMats/src/generics.jl:140
```

### without this PR:
```julia
julia> @which invquad!([2,2], PDiagMat([1,1]), [3 3; 3 3])
invquad!(r::AbstractArray, a::PDiagMat, x::StridedMatrix{T} where T) in PDMats at /home/oxinabox/.julia/packages/PDMats/pIpaj/src/pdiagmat.jl:127

julia> @which invquad!([2,2], PDiagMat([1,1]), @view([3 3; 3 3][:,:]))
invquad!(r::AbstractArray, a::PDiagMat, x::StridedMatrix{T} where T) in PDMats at /home/oxinabox/.julia/packages/PDMats/pIpaj/src/pdiagmat.jl:127

julia> @which invquad!([2,2], PDiagMat([1,1]), @view([3 3; 3 3][[2 4; 1 3]]))
ERROR: no unique matching method found for the specified argument types
```

The only difference I see for this case is that the last one is now longer an error

devmotion:
> The only difference I see for this case is that the last one is now longer an error

Yes, exactly, that's what I meant - it's not an error anymore but it uses the generic fallback instead of the method for diagonal matrices.

oxinabox:
ah right, there is a lot of code here that I think only works for StridedVectors, or rather code that only really works for things indexed by `1:length(x)`.
I think easiest would be to go through and generalize every single method in the package.
Which I am happy to make a follow up PR to do, but i think it would make this PR to large.
I will start that PR now, branching off this one.

src/utils.jl
```diff
@@ -69,11 +69,10 @@ function invwsumsq(w::AbstractVector, a::AbstractVector)
 end
 
 function colwise_dot!(r::AbstractArray, a::AbstractMatrix, b::AbstractMatrix)
-    n = length(r)
-    @check_argdims n == size(a, 2) == size(b, 2) && size(a, 1) == size(b, 1)
-    for j = 1:n
+    @check_argdims size(a, 2) == size(b, 2) && size(a, 1) == size(b, 1)
```
oxinabox, line 72: They have to be using the same axes.
As we index them with the same indicies
```suggestion
    @check_argdims(axes(a) == axes(b))
```

oxinabox:
Given  #163 is open, how do you feel about merging this?
The only behavourally change is to things that errored before.
And we have PRs open that will also make some of those fast.

src/utils.jl
```diff
@@ -69,11 +69,10 @@ function invwsumsq(w::AbstractVector, a::AbstractVector)
...
-    @check_argdims n == size(a, 2) == size(b, 2) && size(a, 1) == size(b, 1)
-    for j = 1:n
+    @check_argdims(axes(a) == axes(b))
+    for j in axes(a, 2)
         v = zero(promote_type(eltype(a), eltype(b)))
-        @simd for i = 1:size(a, 1)
+        @simd for i in axes(a, 1)
             @inbounds v += a[i, j]*b[i, j]
         end
         r[j] = v
```
devmotion, line 78: Similar to my comments in the other PR, `j` is not necessarily a valid index for `r`. We actually want something like
```suggestion
    n = length(r)
    @check_argdims n == size(a, 2) == size(b, 2) && size(a, 1) == size(b, 1)
    for (i, ai, bi) in zip(eachindex(r), eachol(a), eachcol(b))
        r[i] = dot(ai, bi)
```

oxinabox: Per my comment on other PR we know it is because we did `check_argdims(axes(a) == axes(b))`

devmotion: The check `check_argdims(axes(a) == axes(b))` only helps with ensuring that `i, j` are valid indices for `a` and `b`. But it seems it doesn't guarantee that `j` can be used as index for `r`.

src/utils.jl
```diff
@@ -69,11 +69,10 @@ function invwsumsq(w::AbstractVector, a::AbstractVector)
 end
 
 function colwise_dot!(r::AbstractArray, a::AbstractMatrix, b::AbstractMatrix)
-    n = length(r)
-    @check_argdims n == size(a, 2) == size(b, 2) && size(a, 1) == size(b, 1)
-    for j = 1:n
+    @check_argdims(axes(a) == axes(b))
```
oxinabox, line 72:
```suggestion
    @check_argdims(axes(a, 2) == eachindex(r))
```

src/utils.jl
```diff
@@ -69,11 +69,10 @@ function invwsumsq(w::AbstractVector, a::AbstractVector)
 end
 
 function colwise_dot!(r::AbstractArray, a::AbstractMatrix, b::AbstractMatrix)
-    n = length(r)
-    @check_argdims n == size(a, 2) == size(b, 2) && size(a, 1) == size(b, 1)
-    for j = 1:n
+    @check_argdims(axes(a, 2) == eachindex(r))
```
oxinabox, line 72:
```suggestion
    @check_argdims(axes(a) == axes(b))
    @check_argdims(axes(a, 2) == eachindex(r))
```

devmotion (approved):
Looks good to me, thanks!

Any additional comments or thoughts, @andreasnoack?

matbesancon:
Any idea where the nightly tests error comes from?
https://github.com/JuliaStats/PDMats.jl/runs/6593682307?check_suite_focus=true#step:6:185
```
BandedMatrices: Error During Test at /home/runner/work/PDMats.jl/PDMats.jl/test/specialarrays.jl:50
  Got exception outside of a @test
  UndefVarError: liblapack not defined
  Stacktrace:
    [1] pbtrf!(uplo::Char, m::Int64, kd::Int64, A::Matrix{Float64})
      @ BandedMatrices ~/.julia/packages/BandedMatrices/dt29n/src/lapack.jl:210
    [2] banded_chol!
      @ ~/.julia/packages/BandedMatrices/dt29n/src/symbanded/BandedCholesky.jl:4 [inlined]
```

devmotion:
Seems to be caused by the tests with BandedMatrices, so it might not be related to (or rather caused by) PDMats.

devmotion:
The same test errors we've seen for a while on the master branch (eg https://github.com/JuliaStats/PDMats.jl/runs/6388016866?check_suite_focus=true), so it does not seem related to this PR.

oxinabox:
can this please be merged this is blocking some work i want to wrap up

matbesancon:
done, thanks!

PR #163, opened 2022-05-24
Remove StridedMatrix restriction and use modern index handling

oxinabox:
This PR is a follow up to https://github.com/JuliaStats/PDMats.jl/pull/162#issuecomment-1135971969

It relaxes the StridedMatrix restrictions and uses `eachindex`, `axes` and broadcasting rather than `1:size(x,2)` etc.
A lot of places were using `StridedMatrix` with the assumption that it indexed safe with that scheme.
(Which is not infact true, AFAIK, as any array extending `DenseArray` is a `StridedArray`.
And `DenseArray only requires contiguously in memory.
But realistically `OffsetArrays` doesn't actually subtype `DenseArray`, so doesn't really matter)

This PR changes:
 - `1:size(x,2)` to `axes(x, 2)` and `1:length(x)` to `eachindex(x)` (this is done even when the types were not relaxed as it reads better)
 - rewrites some things entirely to just use broadcasting
 - `@checkargs` to now checks `eachindex` and `axes`, rather than `length` and `size`. (this makes certain they are indexed the same when we are using one eachindex rather than other. For `eachindex` we could use the 2 arg form but mostly we want to match a `eachindex(a)` against a `axes(x,2)`, so can't)
 - Relaxes `StridedArray` into `AbstractArray`

It currently adds no tests, I am not familar with how this package is tested so I don't know if there is an easy way to test it e.g. with OffsetArrays or noncontigious SubArrays or something.


ararslan (approved):
Thanks Frames! The implementation looks good to me.

It would be ideal to have tests but the way the package is tested is indeed a bit unusual: it uses the `pdtest_*` functions defined in `test/testutils.jl` to test a number of things at once. Those functions are constrained such that one of the arguments has to be a `Matrix`, which I think would be the first thing that would need to be changed in order to support e.g. OffsetArrays.

src/pdiagmat.jl
```diff
@@ -74,63 +74,35 @@ LinearAlgebra.sqrt(a::PDiagMat) = PDiagMat(map(sqrt, a.diag))
...
 quad(a::PDiagMat, x::AbstractVector) = wsumsq(a.diag, x)
 invquad(a::PDiagMat, x::AbstractVector) = invwsumsq(a.diag, x)
 
-function quad!(r::AbstractArray, a::PDiagMat, x::StridedMatrix)
+function quad!(r::AbstractArray, a::PDiagMat, x::AbstractMatrix)
     m, n = size(x)
     ad = a.diag
     @check_argdims m == length(ad) && length(r) == n
-    @inbounds for j = 1:n
+    @inbounds for j in axes(x, 2)
```
devmotion, line 89: This is not correct I think: We need indices for r but also indices of x along the second dimension. They don't have to match, and in particular r might be of any shape currently. So I think we want something like
```suggestion
    @inbounds for (ir, j) in zip(eachindex(r), axes(x, 2))
```

src/pdiagmat.jl
```diff
@@ -74,63 +74,35 @@ LinearAlgebra.sqrt(a::PDiagMat) = PDiagMat(map(sqrt, a.diag))
...
         s = zero(promote_type(eltype(ad), eltype(x)))
-        for i in 1:m
+        for i in axes(x, 1)
```
devmotion, line 91: Same here, i is possibly invalid for indexing ad.
```suggestion
        for (adi, i) in zip(ad, axes(x, 1))
```

src/pdiagmat.jl
```diff
@@ -74,63 +74,35 @@ LinearAlgebra.sqrt(a::PDiagMat) = PDiagMat(map(sqrt, a.diag))
...
     r
 end
 
-function invquad!(r::AbstractArray, a::PDiagMat, x::StridedMatrix)
+function invquad!(r::AbstractArray, a::PDiagMat, x::AbstractMatrix)
     m, n = size(x)
     ad = a.diag
     @check_argdims m == length(ad) && length(r) == n
-    @inbounds for j = 1:n
+    @inbounds for j in axes(x, 2)
```
devmotion, line 103: Same here.

src/pdiagmat.jl
```diff
@@ -74,63 +74,35 @@ LinearAlgebra.sqrt(a::PDiagMat) = PDiagMat(map(sqrt, a.diag))
...
         s = zero(zero(eltype(x)) / zero(eltype(ad)))
-        for i in 1:m
+        for i in axes(x, 1)
```
devmotion, line 105: And here.

src/pdsparsemat.jl
```diff
@@ -57,31 +58,31 @@ LinearAlgebra.sqrt(A::PDSparseMat) = PDMat(sqrt(Hermitian(Matrix(A))))
 
 ### whiten and unwhiten
 
-function whiten!(r::StridedVecOrMat, a::PDSparseMat, x::StridedVecOrMat)
+function whiten!(r::AbstractVecOrMat, a::PDSparseMat, x::AbstractVecOrMat)
     r[:] = sparse(chol_lower(a.chol)) \ x
```
devmotion, line 61: Could we use an inplace method instead of creating a copy?

src/pdsparsemat.jl
```diff
@@ -57,31 +58,31 @@ LinearAlgebra.sqrt(A::PDSparseMat) = PDMat(sqrt(Hermitian(Matrix(A))))
...
     return r
 end
 
-function unwhiten!(r::StridedVecOrMat, a::PDSparseMat, x::StridedVecOrMat)
+function unwhiten!(r::AbstractVecOrMat, a::PDSparseMat, x::AbstractVecOrMat)
     r[:] = sparse(chol_lower(a.chol)) * x
```
devmotion, line 66: Same here?

src/pdsparsemat.jl
```diff
@@ -57,31 +58,31 @@ LinearAlgebra.sqrt(A::PDSparseMat) = PDMat(sqrt(Hermitian(Matrix(A))))
...
 
-quad(a::PDSparseMat, x::StridedVector) = dot(x, a * x)
-invquad(a::PDSparseMat, x::StridedVector) = dot(x, a \ x)
+quad(a::PDSparseMat, x::AbstractVector) = dot(x, a * x)
+invquad(a::PDSparseMat, x::AbstractVector) = dot(x, a \ x)
 
-function quad!(r::AbstractArray, a::PDSparseMat, x::StridedMatrix)
-    for i in 1:size(x, 2)
+function quad!(r::AbstractArray, a::PDSparseMat, x::AbstractMatrix)
+    for i in axes(x, 2)
```
devmotion, line 78: Also here i might not be a valid index for r.

src/utils.jl
```diff
@@ -12,34 +12,32 @@ _rcopy!(r::StridedVecOrMat, x::StridedVecOrMat) = (r === x || copyto!(r, x); r)
...
             @inbounds r[i] = a[i] + b[i] * c
         end
     end
     return r
 end
 
 function _adddiag!(a::Union{Matrix, SparseMatrixCSC}, v::Real)
-    n = size(a, 1)
-    for i = 1:n
+    for i in axes(a, 1)
```
devmotion, line 27: Maybe just
```suggestion
    for i in diagind(a)
```

src/utils.jl
```diff
@@ -50,19 +48,21 @@ _adddiag(a::Union{Matrix, SparseMatrixCSC}, v::Real) = _adddiag!(copy(a), v)
 _adddiag(a::Union{Matrix, SparseMatrixCSC}, v::Vector, c::Real) = _adddiag!(copy(a), v, c)
 _adddiag(a::Union{Matrix, SparseMatrixCSC}, v::Vector{T}) where {T<:Real} = _adddiag!(copy(a), v, one(T))
 
+
 function wsumsq(w::AbstractVector, a::AbstractVector)
-    @check_argdims(length(a) == length(w))
+    @check_argdims(eachindex(a) == eachindex(w))
```
devmotion, line 53: Maybe we should be less restrictive and just change the loop below instead?

src/utils.jl
```diff
@@ -50,19 +48,21 @@ _adddiag(a::Union{Matrix, SparseMatrixCSC}, v::Real) = _adddiag!(copy(a), v)
...
     s = zero(promote_type(eltype(w), eltype(a)))
-    for i = 1:length(a)
+    for i in eachindex(w)
```
devmotion, line 55: Maybe just
```suggestion
    for (ai, wi) in zip(a, w)
```

src/utils.jl
```diff
@@ -50,19 +48,21 @@ _adddiag(a::Union{Matrix, SparseMatrixCSC}, v::Real) = _adddiag!(copy(a), v)
...
         @inbounds s += abs2(a[i]) * w[i]
     end
     return s
 end
 
 function invwsumsq(w::AbstractVector, a::AbstractVector)
-    @check_argdims(length(a) == length(w))
+    @check_argdims(eachindex(a) == eachindex(w))
```
devmotion, line 62: Same as above.

src/utils.jl
```diff
@@ -82,11 +82,10 @@ function colwise_dot!(r::AbstractArray, a::AbstractMatrix, b::AbstractMatrix)
 end
 
 function colwise_sumsq!(r::AbstractArray, a::AbstractMatrix, c::Real)
-    n = length(r)
-    @check_argdims n == size(a, 2)
-    for j = 1:n
+    @check_argdims(eachindex(r) == axes(a, 2))
```
devmotion, line 85: Same as above, maybe revert this change.

src/utils.jl
```diff
@@ -95,11 +94,10 @@ function colwise_sumsq!(r::AbstractArray, a::AbstractMatrix, c::Real)
 end
 
 function colwise_sumsqinv!(r::AbstractArray, a::AbstractMatrix, c::Real)
-    n = length(r)
-    @check_argdims n == size(a, 2)
-    for j = 1:n
+    @check_argdims(eachindex(r) == axes(a, 2))
```
devmotion, line 97: Here as well.

src/utils.jl
```diff
@@ -95,11 +94,10 @@ function colwise_sumsq!(r::AbstractArray, a::AbstractMatrix, c::Real)
...
+    for j in axes(a, 2)
         v = zero(eltype(a))
-        @simd for i = 1:size(a, 1)
+        @simd for i in axes(a, 1)
             @inbounds v += abs2(a[i, j])
         end
         r[j] = v / c
```
devmotion, line 105: Maybe just change to
for (i, ai) in zip(eachindex(r), eachcol(a))
r[i] = sum(abs2, ai) / c
end
?

src/utils.jl
```diff
@@ -82,11 +82,10 @@ function colwise_dot!(r::AbstractArray, a::AbstractMatrix, b::AbstractMatrix)
...
-    @check_argdims n == size(a, 2)
-    for j = 1:n
+    @check_argdims(eachindex(r) == axes(a, 2))
+    for j in axes(a, 2)
         v = zero(eltype(a))
-        @simd for i = 1:size(a, 1)
+        @simd for i in axes(a, 1)
             @inbounds v += abs2(a[i, j])
         end
         r[j] = v*c
```
devmotion, line 92: Maybe just change to
for (i, ai) in zip(eachindex(r), eachcol(a))
r[i] = sum(abs2, ai) * c
end
?

src/pdsparsemat.jl
oxinabox: I didn't change this code, it is out of scope for the PR

src/utils.jl
oxinabox: This is safe. This makes sure the indexes *are* in-fact the same
This is what I think this PR should do everywhere.
I missed it above.

Making sure that things with different indexes work the same as before would require much more extensive testing.
And i think can be left for another PR.

src/utils.jl
oxinabox: I think out of scope for this PR.
I am avoiding changes to logic except in the one case where I can delete a method entirely.

src/pdiagmat.jl
oxinabox: This is correct as long as I also check the axes (rather than the length) above in the `checkdims`.
It isn't maximally general.
But such generality i think better to come in a follow up PR that does a lot more testing.
Much easier to screw up.

oxinabox: fixed check (just noting here so will push shortly)

src/pdiagmat.jl
oxinabox: fixed check (just noting here so will push shortly)

src/pdiagmat.jl
oxinabox: fixed check (just noting here so will push shortly)

src/pdsparsemat.jl
oxinabox: fixed check

src/utils.jl
oxinabox: Per others, I would like to leave this for a future PR that adds more careful tests to go with such a change

oxinabox:
@devmotion what do you think.
is this good to go now?

Or do you want 
  - Tests of this directly
  - Actually supporting mixing of arrays with different axes

oxinabox:
bump

devmotion (commented):
Sorry for the delay, I was waiting for #162 since I assume (it seems incorrectly) they depend on each other.

> Or do you want
>
> - Tests of this directly
> - Actually supporting mixing of arrays with different axes

I think some of the restrictions could be relaxed but this can be done in follow-up PRs. If you manage to include it, I think additional tests with more general arrays, possibly using non-standard axes, such as OffsetArrays would be very good to add. Tests currently pass, so it seems StridedArrays still work (at least whatever we test for), but it doesn't tell us if it actually works for more general arrays as desired.

Generally, I guess updating the tests would involve adding other array types in lines such as https://github.com/JuliaStats/PDMats.jl/blob/6674103af15f405fdefd7d203369fa8081ca260f/test/testutils.jl#L253 and https://github.com/JuliaStats/PDMats.jl/blob/6674103af15f405fdefd7d203369fa8081ca260f/test/testutils.jl#L302.

matbesancon:
@oxinabox can  you also rebase on master?

src/utils.jl
```diff
@@ -12,34 +12,32 @@ _rcopy!(r::StridedVecOrMat, x::StridedVecOrMat) = (r === x || copyto!(r, x); r)
...
+    for i in diagind(a)
+        @inbounds a[i] += v
     end
     return a
 end
 
 function _adddiag!(a::Union{Matrix, SparseMatrixCSC}, v::Vector, c::Real)
-    n = size(a, 1)
-    @check_argdims length(v) == n
+    @check_argdims eachindex(v) == axes(a, 1)
```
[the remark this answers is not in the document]
oxinabox: This is a stricter test than what was there before, but it is correct

oxinabox:
The approach here of just demanding indexes are the same seems to be what is recommended by this thread.
Rather than assuming that iteration order is order we should add together.
https://discourse.julialang.org/t/discussion-on-why-i-no-longer-recommend-julia-by-yuri-vishnevsky/81151/314?u=oxinabox

matbesancon (approved):

matbesancon:
@devmotion I'll let you have a last look but LGTM

devmotion (approved):
LGTM, thanks!

Let's add more tests in a separate PR 🙂

PR #165, opened 2022-06-02
Improve constructor `PDMat(::Cholesky)`

### LINKED ISSUE #137: PDMat constructor fails for cholesky of a diagonal matrix
A minimal example:

```julia
using PDMats, LinearAlgebra
x = rand(10, 10)
PDMat(cholesky(Diagonal(x'x)))
```

<details>
  <summary>Stacktrace</summary>

```julia
ERROR: MethodError: Cannot `convert` an object of type Matrix{Float64} to an object of type Diagonal{Float64, Vector{Float64}}
Closest candidates are:
  convert(::Type{var"#s832"} where var"#s832"<:Diagonal, ::Union{LinearAlgebra.AbstractTriangular, Bidiagonal, Diagonal, SymTridiagonal, Tridiagonal}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/special.jl:53
  convert(::Type{T}, ::Factorization) where T<:AbstractArray at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/factorization.jl:58
  convert(::Type{T}, ::T) where T<:AbstractArray at abstractarray.jl:14
  ...
Stacktrace:
 [1] PDMat(mat::Matrix{Float64}, chol::Cholesky{Float64, Diagonal{Float64, Vector{Float64}}})
   @ PDMats ~/.config/julia/packages/PDMats/G0Prn/src/pdmat.jl:16
 [2] PDMat(fac::Cholesky{Float64, Diagonal{Float64, Vector{Float64}}})
   @ PDMats ~/.config/julia/packages/PDMats/G0Prn/src/pdmat.jl:21
 [3] top-level scope
   @ REPL[2]:1
```

</details>

Due to attempting to convert a `Matrix` to a `Diagonal`: https://github.com/JuliaStats/PDMats.jl/blob/213c37a2a94e7d14c21b54b2eeded4cb76270488/src/pdmat.jl#L16

### LINKED ISSUE #155: Generalize PDMat constructor from Cholesky
Now that #131 is merged, it's faster to construct a `PDMat` with an `SMatrix` than it is with `Cholesky{T, SMatrix}`. I believe this is because the constructor [here](https://github.com/JuliaStats/PDMats.jl/blob/87277241153cde10fa6ec82086e2af2f050240f4/src/pdmat.jl#L20) creates a `Matrix` then [here](https://github.com/JuliaStats/PDMats.jl/blob/87277241153cde10fa6ec82086e2af2f050240f4/src/pdmat.jl#L16) converts it to an `SMatrix`.

Naively, it seems like this could be improved simply by changing
`PDMat(fac::Cholesky) = PDMat(Matrix(fac), fac)`
to
`PDMat(fac::Cholesky) = PDMat(AbstractMatrix(fac), fac)`
in `pdmat.jl`

devmotion:
This PR fixes #155. This PR fixes #137.

## On the master branch

```julia
julia> using PDMats, StaticArrays, BenchmarkTools, LinearAlgebra

julia> A = SMatrix{3,3}(rand(3, 3));

julia> B = A * A' +  I;

julia> @btime PDMat($(cholesky(B)));
  45.653 ns (1 allocation: 128 bytes)
```

## With this PR

```julia
julia> using PDMats, StaticArrays, BenchmarkTools, LinearAlgebra

julia> A = SMatrix{3,3}(rand(3, 3));

julia> B = A * A' +  I;

julia> @btime PDMat($(cholesky(B)));
  9.706 ns (0 allocations: 0 bytes)
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "PDMats"
 uuid = "90014a1f-27ba-587c-ab20-58faa44d9150"
-version = "0.11.11"
+version = "0.11.12"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/pdmat.jl
+++ b/src/pdmat.jl
@@ -17,7 +17,7 @@ function PDMat(mat::AbstractMatrix,chol::Cholesky{T,S}) where {T,S}
 end
 
 PDMat(mat::AbstractMatrix) = PDMat(mat, cholesky(mat))
-PDMat(fac::Cholesky) = PDMat(Matrix(fac), fac)
+PDMat(fac::Cholesky) = PDMat(AbstractMatrix(fac), fac)
 
 ### Conversion
 Base.convert(::Type{PDMat{T}},         a::PDMat) where {T<:Real} = PDMat(convert(AbstractArray{T}, a.mat))
--- a/test/pdmtypes.jl
+++ b/test/pdmtypes.jl
@@ -111,10 +111,14 @@ using Test
     end
 
     @testset "PDMat from Cholesky decomposition of diagonal matrix (#137)" begin
-        x = rand(10, 10)
-        A = Diagonal(x' * x)
-        M = PDMat(cholesky(A))
-        @test M isa PDMat{Float64, typeof(A)}
-        @test Matrix(M) ≈ A
+        # U'*U where U isa UpperTriangular etc.
+        # requires https://github.com/JuliaLang/julia/pull/33334
+        if VERSION >= v"1.4.0-DEV.286"
+            x = rand(10, 10)
+            A = Diagonal(x' * x)
+            M = PDMat(cholesky(A))
+            @test M isa PDMat{Float64, typeof(A)}
+            @test Matrix(M) ≈ A
+        end
     end
 end
--- a/test/pdmtypes.jl
+++ b/test/pdmtypes.jl
@@ -109,4 +109,12 @@ using Test
             @test x / PDSparseMat(sparse(first(A), 1, 1)) ≈ y
         end
     end
+
+    @testset "PDMat from Cholesky decomposition of diagonal matrix (#137)" begin
+        x = rand(10, 10)
+        A = Diagonal(x' * x)
+        M = PDMat(cholesky(A))
+        @test M isa PDMat{Float64, typeof(A)}
+        @test Matrix(M) ≈ A
+    end
 end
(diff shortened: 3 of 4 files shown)


matbesancon (approved):

matbesancon:
a method was missing for `lmul!` in Julia 1.0, I'd suggest requiring the LTS, it was bound to happen at some point
```
  MethodError: no method matching lmul!(::Adjoint{Float64,UpperTriangular{Float64,Diagonal{Float64,Array{Float64,1}}}}, ::SparseMatrixCSC{Float64,Int64})
```

the LTS fail is unrelated again

devmotion:
> a method was missing for `lmul!` in Julia 1.0, I'd suggest requiring the LTS, it was bound to happen at some point
> 
> ```
>   MethodError: no method matching lmul!(::Adjoint{Float64,UpperTriangular{Float64,Diagonal{Float64,Array{Float64,1}}}}, ::SparseMatrixCSC{Float64,Int64})
> ```

I don't think this warrants an update of the Julia compat entry. This PR is an improvement even for Julia 1.0 (fixes #155 - can be seen e.g. since CI passes for the first commit), just one specific example (namely the one in #137 - test for it added in the second commit) is still broken. However, on Julia <= 1.3 that example fails with the master branch as well due to the same problem:
```julia
julia> using LinearAlgebra

julia> x = rand(10, 10);

julia> A = Diagonal(x' * x);

julia> Matrix(cholesky(A)) # used on the master branch
ERROR: MethodError: no method matching lmul!(::Adjoint{Float64,UpperTriangular{Float64,Diagonal{Float64,Array{Float64,1}}}}, ::SparseArrays.SparseMatrixCSC{Float64,Int64})
...

julia> AbstractMatrix(cholesky(A)) # used in this PR
ERROR: MethodError: no method matching lmul!(::Adjoint{Float64,UpperTriangular{Float64,Diagonal{Float64,Array{Float64,1}}}}, ::SparseArrays.SparseMatrixCSC{Float64,Int64})
...
```
So in my opinion, since already on the master branch that specific example fails due to this bug in Julia itself and hence the underlying issue in PDMats (and its fix here) are not even relevant, it is fine to just not fix that example on Julia <= 1.3 and instead tell users that want to use `cholesky(Diagonal(...))` inputs to update their Julia version.

> the LTS fail is unrelated again

The unrelated test failures occur with Julia nightly (caused by BlockDiagonals), LTS should be fine (but is not part of the CI tests).

PR #166, opened 2022-06-05
Test LTS

devmotion:
I think it would be good to test the new LTS as well.


devmotion:
Test errors with Julia nightly are unrelated, known (exist e.g. also on the master branch), and caused by BandedMatrices. https://github.com/JuliaMatrices/BandedMatrices.jl/pull/258 should fix them.

devmotion:
I'll merge before #165 so we can run tests with 1.6 there as well.

Issue #140, opened 2021-10-27
Design question/discussion: PDMat without constructing full matrix


st--:
As pointed out in the README, `PDMat` always stores both Cholesky factorisation and full matrix. For many operations that are commonly used downstream, such as logdet and the quadratic forms, the full matrix is never used. So when the `PDMat` is constructed from an already known Cholesky factor (e.g. because we optimize for a variationally optimal covariance matrix), this is an unnecessary computation (see https://github.com/invenia/ParameterHandling.jl/issues/41#issuecomment-936495430).

What are the design reasons for the current setup? What would it take to change this? Could we, for example, have some lazy way of computing `mat` (or `chol`) only when it is actually required [and then caching it]?

st--:
What are the PDMats.jl maintainers' thoughts on this?

devmotion:
Was fixed by https://github.com/JuliaStats/PDMats.jl/pull/168: `Cholesky` supports the PDMats API, so if you do not need/want the full matrix, you can directly operate with `Cholesky`.

PR #168, opened 2022-06-22
Define functions for `Cholesky`

devmotion:
This PR defines

 - `dim`
 - `whiten`, `whiten!`
 - `unwhiten`, `unwhiten!`
 - `quad`, `quad!`
 - `invquad`, `invquad!`
 - `X_A_Xt`, `Xt_A_X`, `X_invA_Xt`, `Xt_invA_X`

for `Cholesky`.

Ultimately, I think we should forward (some of) the generic fallbacks for `AbstractMatrix` to these functions (I noticed they could also be cleaned a bit since e.g. https://github.com/JuliaStats/PDMats.jl/blob/8d88c15f7c9931038ab4b741fea92f7fc990da4d/src/generics.jl#L124 is never hit for `x::AbstractMatrix` due to the definition in https://github.com/JuliaStats/PDMats.jl/blob/8d88c15f7c9931038ab4b741fea92f7fc990da4d/src/generics.jl#L125; additionally, it is problematic e.g. for `StaticArray`s that out-of-place methods fall back to in-place methods). However, I assumed it would be cleaner to not touch the dispatch pipeline in this PR but only add and test definitions for `Cholesky`.

This PR also provides a solution to https://github.com/JuliaStats/PDMats.jl/issues/140 @st--: If one does not want to construct the full matrix, it is possible to work with `Cholesky` directly (at least when using the PDMats-specific API).

### DIFF
--- a/test/chol.jl
+++ b/test/chol.jl
@@ -2,19 +2,28 @@ using LinearAlgebra, PDMats
 using PDMats: chol_lower, chol_upper
 
 @testset "chol_lower" begin
-    A = rand(100, 100)
+    d = 100
+    A = rand(d, d)
     C = A'A
+    invC = inv(C)
     size_of_one_copy = sizeof(C)
-    @assert size_of_one_copy > 100  # ensure the matrix is large enough that few-byte allocations don't matter
+    @assert size_of_one_copy > d  # ensure the matrix is large enough that few-byte allocations don't matter
 
     chol_lower(C)
     @test (@allocated chol_lower(C)) < 1.05 * size_of_one_copy  # allow 5% overhead
 
+    X = randn(d, 10)
     for uplo in (:L, :U)
         ch = cholesky(Symmetric(C, uplo))
         chol_lower(ch)
         @test (@allocated chol_lower(ch)) < 33  # allow small overhead for wrapper types
         chol_upper(ch)
         @test (@allocated chol_upper(ch)) < 33  # allow small overhead for wrapper types
+
+        # Only test dim, `quad`/`invquad`, `whiten`/`unwhiten`, and tri products
+        @test dim(ch) == size(C, 1)
+        pdtest_quad(ch, C, invC, X, 0)
+        pdtest_triprod(ch, C, invC, X, 0)
+        pdtest_whiten(ch, C, 0)
     end
 end
--- a/test/chol.jl
+++ b/test/chol.jl
@@ -13,6 +13,7 @@ using PDMats: chol_lower, chol_upper
         @test (@allocated chol_lower(C)) < 1.05 * size_of_one_copy  # allow 5% overhead
         @test (@allocated chol_upper(C)) < 1.05 * size_of_one_copy
 
+        X = randn(d, 10)
         for uplo in (:L, :U)
             ch = cholesky(Symmetric(C, uplo))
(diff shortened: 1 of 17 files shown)


btmit:
It might be worth extending addition as well. Below uses a full rank update to add two matrices directly using their Cholesky decompositions. IIRC this is more numerically stable as well. Unfortunately this doesn't work for `StaticArrays`   #https://github.com/JuliaArrays/StaticArrays.jl/issues/930

``` julia
using LinearAlgebra, PDMats, BenchmarkTools

function fullrankupdate(c1::Cholesky, c2::Cholesky)
    cout = copy(c1)
    return fullrankupdate!(cout, c2)
end

function fullrankupdate!(c1::Cholesky, c2::Cholesky)
    for c in eachcol(PDMats.chol_lower(c2))
        lowrankupdate!(c1, c)
    end
    return c1
end

c1 = cholesky([1.0 0.5; 0.5 2.0])
c2 = cholesky([2.0 0.2; 0.2 1.0])

@btime PDMat($c1) + PDMat($c2);

@btime fullrankupdate($c1, $c2);

@assert PDMats.chol_lower(cholesky(PDMat(c1) + PDMat(c2))) ≈ PDMats.chol_lower(fullrankupdate(c1, c2))
```
>     458.122 ns (12 allocations: 688 bytes)
>     70.585 ns (5 allocations: 240 bytes)
>     true

devmotion:
I don't think this should be done in this PR. In general, I'm not sure if it's a good idea at all to define addition for `Cholesky` at all but if it is desired it should probably be done in base.

matbesancon (approved):

test/specialarrays.jl
```diff
@@ -52,6 +55,33 @@ using StaticArrays
             @test Xt_invA_X(A, Y) isa SMatrix{10, 10, Float64}
             @test Xt_invA_X(A, Y) ≈ Matrix(Y)' * (Matrix(A) \ Matrix(Y))
         end
+
+        # TODO: Fix for `PDMat` and `PDiagMat`
```
matbesancon, line 59: is there an issue tracking this TODO?

devmotion: No, no issue. But it will be fixed by https://github.com/JuliaStats/PDMats.jl/pull/183 🙂

test/chol.jl
```diff
@@ -5,18 +5,26 @@ using PDMats: chol_lower, chol_upper
...
         C = A'A
+        invC = inv(C)
         size_of_one_copy = sizeof(C)
         @assert size_of_one_copy > 100  # ensure the matrix is large enough that few-byte allocations don't matter
 
         @test chol_lower(C) ≈ chol_upper(C)'
         @test (@allocated chol_lower(C)) < 1.05 * size_of_one_copy  # allow 5% overhead
         @test (@allocated chol_upper(C)) < 1.05 * size_of_one_copy
 
+        X = randn(d, 10)
```
andreasnoack, line 16: What is this one?

devmotion: `d` should be 100 and used above as well. It's a remaining (final?) issue from the merge commit https://github.com/JuliaStats/PDMats.jl/pull/168/commits/afaf6999a7c6d6a55042b3b6a5342768faf75600 that broke tests a bit.

andreasnoack (approved):

PR #171, opened 2022-07-08
dim(a) -> LinearAlgebra.checksquare(a)

mohamed82008:
It seems one use of `dim` was not removed from the package when it was deprecated.


andreasnoack (approved):

andreasnoack:
Maybe we should fail when hitting deprecated methods in CI.

PR #175, opened 2023-07-31
Account for pivoting in sparse Cholesky decompositions

### LINKED ISSUE #120: No pivoting for CHOLMOD factors
`whiten!` and `unwhiten` use [`chol_lower` to extract the lower triangular component of Cholesky decompositions](https://github.com/JuliaStats/PDMats.jl/blob/9fe6b3cb4f5cb256405940a42773d4eddde25d1a/src/chol.jl#L7). For `SuiteSparse.CHOLMOD.Factor{T}`, this extracts the lower triangular factor, but does not account for pivoting. This results in the same problem seen [on this Discourse thread](https://discourse.julialang.org/t/getting-strange-mv-normal-samples-based-on-sparse-precision-matrix/48001).

Using the sparse precision matrix attached to that thread,

```julia
using DataFrames, CSV

q_df = CSV.read("precision_matrix.csv")
point_df = CSV.read("mesh_points.csv")
Q = sparse(q_df.i, q_df.j, q_df.q)
```

Generate a realization,

```julia
using Random, SparseArrays, LinearAlgebra
using Distributions
using PDMats

Random.seed!(2)
z = randn(size(Q, 1))

pdQ = PDSparseMat(cholesky(Q))
xldiv = pdQ.chol.UP \ z
xwh = whiten(pdQ, z)
```
These two realizations should be the same, but they are not.

```julia
julia> maximum(xldiv - xwh)
45.00349329625478
```

This is also clear when they are plotted.

```julia
using Plots
pldiv = surface(point_df.x, point_df.y, xqs, camera=(45, 80), title="ldiv")
pwh = surface(point_df.x, point_df.y, xwh, camera=(45, 80), title="whiten")
plot(pldiv, pwh, layout=(1,2), size=(1200, 400), colorbar=false)
```

The `ldiv` version results in a random field with the expected structure, while the `whiten` version does not.

comparison

Downstream, this means that at least `rand(::Distributions.MvNormalCanon)` is giving incorrect results. I haven't looked closely at other methods in Distributions.jl yet.

devmotion:
As explained in the docstring, one should be careful when working with Cholesky decompositions of sparse arrays since the property `:L` does not take into account the pivoting - in contrast to `:PtL` which also includes the necessary permutations.

This PR changes `chol_lower` and `chol_upper` for such sparse Cholesky decompositions such that pivoting is taken into account. Unfortunately, since SparseArrays does not support multiplication of such `FactorComponent`s, code that multiplies the output of `chol_lower`/`chol_upper` requires a different, explicitly constructed sparse equivalent of these factors. Alternatively, one could define internal functions for multiplying the output of `chol_lower`/`chol_upper`, let them fall back to `*` by default, and add special implementations for these sparse factors. I've not done this for now because I was not completely sure whether the sparse special case justifies such changes to the (internal) API.

Additionally, the PR adds a seemingly missing definition of `chol_upper(::Matrix)` and fixes a few test errors in test/pdmtypes.jl and test/specialarrays.jl caused by upstream bugs.

Fixes #120.

Edit: I suggest merging #180 first. That PR fixes the test issues on the master branch properly whereas this PR here only contains a workaround.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "PDMats"
 uuid = "90014a1f-27ba-587c-ab20-58faa44d9150"
-version = "0.11.18"
+version = "0.11.20"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "PDMats"
 uuid = "90014a1f-27ba-587c-ab20-58faa44d9150"
-version = "0.11.17"
+version = "0.11.18"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/chol.jl
+++ b/src/chol.jl
@@ -15,9 +15,12 @@ chol_lower(a::Matrix) = cholesky(Symmetric(a, :L)).L
 # NOTE: Formally, the line above should use Hermitian() instead of Symmetric(),
 # but this currently has an AutoDiff issue in Zygote.jl, and PDMat is
 # type-restricted to be Real, so they are equivalent.
+chol_upper(a::Matrix) = cholesky(Symmetric(a, :U)).U
 
 if HAVE_CHOLMOD
     CholTypeSparse{T} = SuiteSparse.CHOLMOD.Factor{T}
 
-    chol_lower(cf::CholTypeSparse) = cf.L
+    # Take into account pivoting!
+    chol_lower(cf::CholTypeSparse) = cf.PtL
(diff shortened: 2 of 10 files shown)


devmotion:
Test failures on Julia 1.6 seem to be caused by an upstream issue in ArrayLayouts: https://github.com/JuliaLinearAlgebra/ArrayLayouts.jl/pull/161

devmotion:
Maybe, I'm not sure. It would require a breaking change though so it's also unclear whether/how quickly the package ecosystem would adapt the breaking release.

In any case, I think this consideration is unrelated to the PR. The PR just fixes bugs in the current release of PDMats.

devmotion:
This PR is ready for review as well @andreasnoack 🙂

andreasnoack (approved):

PR #178, opened 2023-09-20
Redundant dimension wastes space and allows disagreement

chriselrod:
Tests also fail on Julia 1.9 on the master branch.
I fixed one of the failures, but the BandedMatrices failure is obviously unrelated so I figured someone else could handle it in a separate PR.

This PR should be non-breaking. I left the old constructors, and added `getproperty` and `propertynames` methods so that the old behaviors are preserved.

Thus, I think bumping the patch version to `0.11.18` should be fine, but I could make it `0.12.0` instead.

### DIFF
--- a/src/pdiagmat.jl
+++ b/src/pdiagmat.jl
@@ -4,12 +4,7 @@ Positive definite diagonal matrix.
 struct PDiagMat{T<:Real,V<:AbstractVector{T}} <: AbstractPDMat{T}
     diag::V
 end
-function PDiagMat(dim::Int, diag::V) where {T,V<:AbstractVector{T}}
-    if length(diag) != dim
-        throw(DimensionMismatch("Dimensions of diag and dim are inconsistent."))
-    end
-    return PDiagMat{T,V}(diag)
-end
+@deprecate PDiagMat(dim, diag) PDiagMat(diag)
 
 function Base.getproperty(a::PDiagMat, s::Symbol)
     if s === :dim
--- a/src/pdiagmat.jl
+++ b/src/pdiagmat.jl
@@ -64,8 +64,12 @@ function \(a::PDiagMat, x::AbstractVecOrMat)
 end
 function /(x::AbstractVecOrMat, a::PDiagMat)
     @check_argdims a.dim == size(x, 2)
-    # return matrix for 1-element vectors `x`, consistent with LinearAlgebra
-    return reshape(x, Val(2)) ./ permutedims(a.diag) # = (a' \ x')'
+    if VERSION < v"1.9-"
+        # return matrix for 1-element vectors `x`, consistent with LinearAlgebra < 1.9
+        return reshape(x, Val(2)) ./ permutedims(a.diag) # = (a' \ x')'
+    else
+        return x ./ (x isa AbstractVector ? a.diag : a.diag')
+    end
 end
 Base.kron(A::PDiagMat, B::PDiagMat) = PDiagMat(vec(permutedims(A.diag) .* B.diag))
 
--- a/src/pdiagmat.jl
+++ b/src/pdiagmat.jl
@@ -4,9 +4,17 @@ Positive definite diagonal matrix.
 struct PDiagMat{T<:Real,V<:AbstractVector{T}} <: AbstractPDMat{T}
     diag::V
 end
+function PDiagMat(dim::Int, diag::V) where {T,V<:AbstractVector{T}}
+    if length(diag) != dim
+        throw(DimensionMismatch("Dimensions of diag and dim are inconsistent."))
+    end
+    return PDiagMat{T,V}(diag)
+end
 
 function Base.getproperty(a::PDiagMat, s::Symbol)
-    s === :dim && return length(getfield(a, :diag))
+    if s === :dim
+        return length(getfield(a, :diag))
+    end
     return getfield(a, s)
 end
 Base.propertynames(::PDiagMat) = (:diag, :dim)
--- a/src/pdiagmat.jl
+++ b/src/pdiagmat.jl
@@ -64,8 +64,7 @@ function \(a::PDiagMat, x::AbstractVecOrMat)
 end
 function /(x::AbstractVecOrMat, a::PDiagMat)
     @check_argdims a.dim == size(x, 2)
-    (VERSION >= v"1.9" && x isa AbstractVector) && return x ./ a.diag
-    # return matrix for 1-element vectors `x`, consistent with LinearAlgebra on old Julia
+    # return matrix for 1-element vectors `x`, consistent with LinearAlgebra
     return reshape(x, Val(2)) ./ permutedims(a.diag) # = (a' \ x')'
 end
 Base.kron(A::PDiagMat, B::PDiagMat) = PDiagMat(vec(permutedims(A.diag) .* B.diag))
--- a/src/pdiagmat.jl
+++ b/src/pdiagmat.jl
@@ -4,7 +4,6 @@ Positive definite diagonal matrix.
 struct PDiagMat{T<:Real,V<:AbstractVector{T}} <: AbstractPDMat{T}
     diag::V
 end
-@deprecate PDiagMat(dim, diag) PDiagMat(diag)
 
 function Base.getproperty(a::PDiagMat, s::Symbol)
     if s === :dim
--- a/src/pdiagmat.jl
+++ b/src/pdiagmat.jl
@@ -2,11 +2,14 @@
 Positive definite diagonal matrix.
 """
 struct PDiagMat{T<:Real,V<:AbstractVector{T}} <: AbstractPDMat{T}
-    dim::Int
     diag::V
 end
 
-PDiagMat(v::AbstractVector{<:Real}) = PDiagMat{eltype(v),typeof(v)}(length(v), v)
+function Base.getproperty(a::PDiagMat, s::Symbol)
+    s === :dim && return length(getfield(a, :diag))
+    return getfield(a, s)
+end
+Base.propertynames(::PDiagMat) = (:diag, :dim)
 
 AbstractPDMat(A::Diagonal{<:Real}) = PDiagMat(A.diag)
 AbstractPDMat(A::Symmetric{<:Real,<:Diagonal{<:Real}}) = PDiagMat(A.data.diag)
@@ -58,7 +61,8 @@ function \(a::PDiagMat, x::AbstractVecOrMat)
 end
 function /(x::AbstractVecOrMat, a::PDiagMat)
     @check_argdims a.dim == size(x, 2)
-    # return matrix for 1-element vectors `x`, consistent with LinearAlgebra
+    (VERSION >= v"1.9" && x isa AbstractVector) && return x ./ a.diag
+    # return matrix for 1-element vectors `x`, consistent with LinearAlgebra on old Julia
     return reshape(x, Val(2)) ./ permutedims(a.diag) # = (a' \ x')'
 end
 Base.kron(A::PDiagMat, B::PDiagMat) = PDiagMat(vec(permutedims(A.diag) .* B.diag))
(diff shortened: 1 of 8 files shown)


src/pdiagmat.jl
```diff
@@ -2,11 +2,14 @@
 Positive definite diagonal matrix.
 """
 struct PDiagMat{T<:Real,V<:AbstractVector{T}} <: AbstractPDMat{T}
-    dim::Int
     diag::V
 end
 
-PDiagMat(v::AbstractVector{<:Real}) = PDiagMat{eltype(v),typeof(v)}(length(v), v)
+function Base.getproperty(a::PDiagMat, s::Symbol)
+    s === :dim && return length(getfield(a, :diag))
```
devmotion, line 9: I think I'd prefer
```julia
    if s === :dim
        return length(getfield(a, :diag))
    else
        ...
    end
```
It's mostly a matter of personal preference but it also had happened to me that the `.... && ...` form masked a code coverage loss in cases where the branch was not tested at all.

src/pdiagmat.jl
```diff
@@ -58,7 +61,8 @@ function \(a::PDiagMat, x::AbstractVecOrMat)
 end
 function /(x::AbstractVecOrMat, a::PDiagMat)
     @check_argdims a.dim == size(x, 2)
-    # return matrix for 1-element vectors `x`, consistent with LinearAlgebra
+    (VERSION >= v"1.9" && x isa AbstractVector) && return x ./ a.diag
```
devmotion, line 64: I thought there was a longer discussion about this change and it was decided to revert it in the end? Or maybe that was just something similar...

devmotion: In any case, could we improve the check and figure out the exact commit, and its version number, which changed this behaviour?

src/pdmat.jl
```diff
@@ -2,23 +2,32 @@
...
 """
 struct PDMat{T<:Real,S<:AbstractMatrix} <: AbstractPDMat{T}
-    dim::Int
     mat::S
     chol::Cholesky{T,S}
 
-    PDMat{T,S}(d::Int,m::AbstractMatrix{T},c::Cholesky{T,S}) where {T,S} = new{T,S}(d,m,c)
+    PDMat{T,S}(m::AbstractMatrix{T},c::Cholesky{T,S}) where {T,S} = new{T,S}(m,c)
+    function PDMat{T,S}(d::Int, m::AbstractMatrix{T},c::Cholesky{T,S}) where {T,S}
+        LinearAlgebra.checksquare(m) == d || throw(DimensionMismatch("Dimensions of mat and chol are inconsistent."))
```
devmotion, line 10: The error message is not really correct since we don't check the dimension of `chol` at all?

src/pdmat.jl
```diff
@@ -2,23 +2,32 @@
 Full positive definite matrix together with a Cholesky factorization object.
 """
 struct PDMat{T<:Real,S<:AbstractMatrix} <: AbstractPDMat{T}
-    dim::Int
     mat::S
     chol::Cholesky{T,S}
 
-    PDMat{T,S}(d::Int,m::AbstractMatrix{T},c::Cholesky{T,S}) where {T,S} = new{T,S}(d,m,c)
+    PDMat{T,S}(m::AbstractMatrix{T},c::Cholesky{T,S}) where {T,S} = new{T,S}(m,c)
+    function PDMat{T,S}(d::Int, m::AbstractMatrix{T},c::Cholesky{T,S}) where {T,S}
```
devmotion, line 9: Can we deprecate this method?

src/pdiagmat.jl
```diff
@@ -2,11 +2,14 @@
 Positive definite diagonal matrix.
 """
 struct PDiagMat{T<:Real,V<:AbstractVector{T}} <: AbstractPDMat{T}
-    dim::Int
     diag::V
```
devmotion, line 5: Just removing the field means the two-argument inner constructor is broken? Can we add it but deprecate it?

src/pdmat.jl
```diff
@@ -2,23 +2,32 @@
...
         throw(DimensionMismatch("Dimensions of mat and chol are inconsistent."))
-    PDMat{T,S}(d, convert(S, mat), chol)
+    PDMat{T,S}(convert(S, mat), chol)
 end
 
 PDMat(mat::AbstractMatrix) = PDMat(mat, cholesky(mat))
 PDMat(fac::Cholesky) = PDMat(AbstractMatrix(fac), fac)
 
+function Base.getproperty(a::PDMat, s::Symbol)
+    s === :dim && return size(getfield(a, :mat), 1)
```
devmotion, line 26: Similar to above, I'd prefer `if ... else ... end`.

src/pdmat.jl
```diff
@@ -49,9 +58,17 @@ end
 *(a::PDMat, x::AbstractVector) = a.mat * x
 *(a::PDMat, x::AbstractMatrix) = a.mat * x
 \(a::PDMat, x::AbstractVecOrMat) = a.chol \ x
-# return matrix for 1-element vectors `x`, consistent with LinearAlgebra
-/(x::AbstractVecOrMat, a::PDMat) = reshape(x, Val(2)) / a.chol
-
+function /(x::AbstractVecOrMat, a::PDMat)
+    if VERSION >= v"1.9" && x isa AbstractVector
```
devmotion, line 62: Can you use the exact commit?

src/pdsparsemat.jl
```diff
@@ -2,24 +2,34 @@
 Sparse positive definite matrix together with a Cholesky factorization object.
 """
 struct PDSparseMat{T<:Real,S<:AbstractSparseMatrix} <: AbstractPDMat{T}
-    dim::Int
     mat::S
     chol::CholTypeSparse
 
-    PDSparseMat{T,S}(d::Int,m::AbstractSparseMatrix{T},c::CholTypeSparse) where {T,S} =
-        new{T,S}(d,m,c) #add {T} to CholTypeSparse argument once #14076 is implemented
+    function PDSparseMat{T,S}(d::Int,m::AbstractSparseMatrix{T},c::CholTypeSparse) where {T,S}
```
devmotion, line 8: Can we deprecate this constructor?

src/pdsparsemat.jl
```diff
@@ -2,24 +2,34 @@
...
+        LinearAlgebra.checksquare(m) == d ||
+            throw(DimensionMismatch("Dimensions of mat and chol are inconsistent."))
```
devmotion, line 10: Can we make the error message more accurate and actually check the size of `chol` as well?

src/pdsparsemat.jl
```diff
@@ -2,24 +2,34 @@
...
         throw(DimensionMismatch("Dimensions of mat and chol are inconsistent."))
-    PDSparseMat{eltype(mat),typeof(mat)}(d, mat, chol)
+    PDSparseMat{eltype(mat),typeof(mat)}(mat, chol)
 end
 
 PDSparseMat(mat::SparseMatrixCSC) = PDSparseMat(mat, cholesky(mat))
 PDSparseMat(fac::CholTypeSparse) = PDSparseMat(sparse(fac), fac)
 
+function Base.getproperty(a::PDSparseMat, s::Symbol)
+    s == :dim && return size(getfield(a, :mat), 1)
```
devmotion, line 28: Same as above.

src/scalmat.jl
```diff
@@ -53,6 +53,7 @@ function \(a::ScalMat, x::AbstractVecOrMat)
 end
 function /(x::AbstractVecOrMat, a::ScalMat)
     @check_argdims a.dim == size(x, 2)
+    VERSION >= v"1.9" && return x .* inv(a.value)
```
devmotion, line 56: Same as above, use the exact version number?

devmotion:
> Tests also fail on Julia 1.9 on the master branch.

Interesting, tests passed in #175 end of Julia, apart from a bug in ArrayLayouts that I fixed upstream.

src/pdiagmat.jl
chriselrod: I wasn't involved in/didn't see those discussions, but this is what I observed when testing on Julia 1.9.3.

src/pdiagmat.jl
chriselrod: Can I suggest adding a `.JuliaFormatter.toml` to enforce style suggestions, so that we don't have to bother discussing such minutiae on PRs?

src/pdmat.jl
chriselrod: I tend to think of deprecations as breaking.

src/pdiagmat.jl
devmotion: As I said, I think it would be better for coverage analysis. I'd prefer a JuliaFormatter setting (even though I'm not sure if it would help here) but unfortuantely I expect that one has to deal with such questions in JuliaStats PRs for the foreseeable future. Discussions around JuliaFormatter haven't lead to any consensus yet, seems people have strong opinions about formatting code 😅

src/pdiagmat.jl
devmotion: IIRC we use exact commits in most (all?) other such branches in PDMats.

BTW I had https://github.com/JuliaLang/julia/pull/49915 and the linked PR in mind.

src/pdmat.jl
devmotion: IIRC historically in PDMats deprecations were considered non-breaking and removed in one of the following breaking releases.

src/pdiagmat.jl
chriselrod: My strongest opinion on formatting is that we should have `.JuliaFormatter.toml`s almost irrespective of what those settings are, to (a) enforce consistency and (b) avoid these discussions and debates. Bike shedding should be called out for the waste of time it is.
Anyway, I did convert these branches.

Maybe I missed it, but I unfortunately didn't see a setting for this
https://domluna.github.io/JuliaFormatter.jl/dev/

chriselrod: https://github.com/domluna/JuliaFormatter.jl/issues/759

src/pdiagmat.jl
chriselrod: Okay, I reverted this change to bring it into compliance.

src/pdmat.jl
chriselrod: Okay, I've added deprecations.

src/pdmat.jl
chriselrod: Reverted for compliance.

src/scalmat.jl
chriselrod: Reverted for compliance.

chriselrod:
Good to merge?

devmotion:
Looks good but I would like the test errors and inconsistencies on the master branch to be fixed first since tests on Julia 1.9 (and the nightlies) in this PR are aborted due to these problems. I opened #180.

devmotion:
Looks good, and tests pass as well with the recent changes on the master branch 🙂 Speaking about tests, I guess the only thing we should check before merging the PR: Are there existing tests of the `getproperty` implementations and internal constructors? And do we use the deprecated constructors anywhere in our codebase (maybe in `convert`)?

chriselrod:
Re pre-existing `getproperty` tests, only for `.diag`:
```sh
> rg "\.d"
pdmtypes.jl
52:        @test convert(PDiagMat{Float64}, PDiagMat(m)).diag == PDiagMat(convert(Array{Float64}, m)).diag
53:        @test convert(AbstractArray{Float64}, PDiagMat(m)).diag == PDiagMat(convert(Array{Float64}, m)).diag
84:        @test d.diag === v
```

While `Base.size`'s implementation uses `getproperty(a, :dim)`, so that branch should be taken here.

Re deprecated constructors: tests still pass if we delete them, so they aren't used anywhere that is tested.

devmotion (approved):
Thank you @chriselrod! I added a missing deprecation and added tests (of the deprecations, properties, and dimension checks).

PR #181, opened 2023-09-26
Support StaticArrays in X(t)_(inv)A_X(t) with ScalMat

devmotion:
Currently, `ScalMat` does not support `X_A_Xt` etc. with static arrays. For instance, running the tests in this PR on the master branch yields
```julia
StaticArrays: Error During Test at REPL[10]:42
  Test threw exception
  Expression: X_A_Xt(A, X) isa SMatrix{10, 10, Float64}
  setindex!(::SMatrix{10, 10, Float64, 100}, value, ::Int) is not defined.
   Hint: Use `MArray` or `SizedArray` to create a mutable static array
...
StaticArrays: Error During Test at REPL[10]:43
  Test threw exception
  Expression: X_A_Xt(A, X) ≈ Matrix(X) * Matrix(A) * (Matrix(X))'
  setindex!(::SMatrix{10, 10, Float64, 100}, value, ::Int) is not defined.
   Hint: Use `MArray` or `SizedArray` to create a mutable static array
...
StaticArrays: Error During Test at REPL[10]:45
  Test threw exception
  Expression: X_invA_Xt(A, X) isa SMatrix{10, 10, Float64}
  setindex!(::SMatrix{10, 10, Float64, 100}, value, ::Int) is not defined.
   Hint: Use `MArray` or `SizedArray` to create a mutable static array
...
StaticArrays: Error During Test at REPL[10]:46
  Test threw exception
  Expression: X_invA_Xt(A, X) ≈ Matrix(X) * (Matrix(A) \ (Matrix(X))')
  setindex!(::SMatrix{10, 10, Float64, 100}, value, ::Int) is not defined.
   Hint: Use `MArray` or `SizedArray` to create a mutable static array
...
StaticArrays: Error During Test at REPL[10]:48
  Test threw exception
  Expression: Xt_A_X(A, Y) isa SMatrix{10, 10, Float64}
  setindex!(::SMatrix{10, 10, Float64, 100}, value, ::Int) is not defined.
   Hint: Use `MArray` or `SizedArray` to create a mutable static array
...
StaticArrays: Error During Test at REPL[10]:49
  Test threw exception
  Expression: Xt_A_X(A, Y) ≈ (Matrix(Y))' * Matrix(A) * Matrix(Y)
  setindex!(::SMatrix{10, 10, Float64, 100}, value, ::Int) is not defined.
   Hint: Use `MArray` or `SizedArray` to create a mutable static array
...
StaticArrays: Error During Test at REPL[10]:51
  Test threw exception
[...]

PR #183, opened 2023-09-26
Support `(un)whiten` and `(inv)quad` with static arrays

devmotion:
Based on #181, this PR also adds support for `whiten`, `unwhiten`, `quad`, and `invquad` with static arrays.

The PR grew a bit larger, hence I kept it separate from #181 for now.

Main points:
- Do not use mutation in non-mutating methods (in a few cases, allocations could be reduced by specializing on `Array`, see my comment in #181 - should we do this?)
- Improve composability by forwarding calls with generic matrices to `f(..., AbstractPDMat(x), ...)` (as a nice side effect, this means that more efficient implementations are called automatically whenever `AbstractPDMat(x)` is specialized; this is the case, e.g., for `Diagonal` matrices).
- Check dimensions.

### DIFF
--- a/src/generics.jl
+++ b/src/generics.jl
@@ -33,19 +33,6 @@ LinearAlgebra.checksquare(a::AbstractPDMat) = size(a, 1)
 
 ## whiten and unwhiten
 
-whiten!(a::AbstractMatrix, x::AbstractVecOrMat) = whiten!(x, a, x)
-unwhiten!(a::AbstractMatrix, x::AbstractVecOrMat) = unwhiten!(x, a, x)
-
-function whiten!(r::AbstractVecOrMat, a::AbstractMatrix, x::AbstractVecOrMat)
-    v = _rcopy!(r, x)
-    ldiv!(chol_lower(cholesky(a)), v)
-end
-
-function unwhiten!(r::AbstractVecOrMat, a::AbstractMatrix, x::AbstractVecOrMat)
-    v = _rcopy!(r, x)
-    lmul!(chol_lower(cholesky(a)), v)
-end
-
 """
     whiten(a::AbstractMatrix, x::AbstractVecOrMat)
     unwhiten(a::AbstractMatrix, x::AbstractVecOrMat)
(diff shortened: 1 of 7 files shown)


src/pdiagmat.jl
```diff
@@ -86,45 +86,48 @@ LinearAlgebra.sqrt(a::PDiagMat) = PDiagMat(map(sqrt, a.diag))
...
-quad(a::PDiagMat, x::AbstractVector) = wsumsq(a.diag, x)
-invquad(a::PDiagMat, x::AbstractVector) = invwsumsq(a.diag, x)
+function quad(a::PDiagMat, x::AbstractVecOrMat)
+    @check_argdims a.dim == size(x, 1)
+    if x isa AbstractVector
+        return wsumsq(a.diag, x)
+    else
+        # map(Base.Fix1(invquad, a), eachcol(x)) or similar alternatives
+        # do NOT return a `SVector` for inputs `x::SMatrix`.
+        return vec(sum(abs2.(x) .* a.diag; dims = 1))
```
devmotion, line 118: This is a bit unsatisfying - is there any way we could avoid unnecessary allocations but still make StaticArray return the expected types?

src/pdiagmat.jl
```diff
@@ -86,45 +86,48 @@ LinearAlgebra.sqrt(a::PDiagMat) = PDiagMat(map(sqrt, a.diag))
...
-unwhiten!(r::AbstractVecOrMat, a::PDiagMat, x::AbstractVecOrMat) = r .= x .* sqrt.(a.diag)
-
-
 ### quadratic forms
 
-quad(a::PDiagMat, x::AbstractVector) = wsumsq(a.diag, x)
-invquad(a::PDiagMat, x::AbstractVector) = invwsumsq(a.diag, x)
+function quad(a::PDiagMat, x::AbstractVecOrMat)
+    @check_argdims a.dim == size(x, 1)
+    if x isa AbstractVector
```
devmotion, line 113: Not sure if it is helpful/desirable here, but generally I tried to reduce the number of methods to lower the probability for method ambiguity errors. Maybe it's mostly useful for the generic code path in src/generics.jl.

src/pdsparsemat.jl
```diff
@@ -70,20 +70,41 @@ LinearAlgebra.sqrt(A::PDSparseMat) = PDMat(sqrt(Hermitian(Matrix(A))))
 ### whiten and unwhiten
 
 function whiten!(r::AbstractVecOrMat, a::PDSparseMat, x::AbstractVecOrMat)
-    r[:] = sparse(chol_lower(a.chol)) \ x
-    return r
+    @check_argdims axes(r) == axes(x)
+    @check_argdims a.dim == size(x, 1)
+    # ldiv! is not defined for SparseMatrixCSC
```
devmotion, line 75: Sparse arrays are annoying... (but at least with #175 the results will become correct).

src/pdiagmat.jl
devmotion: I added a version with reduced allocations specialized for `Array` arguments.

andreasnoack (commented):
Mostly looks good to me. Just curious about the type parameter restriction.

src/generics.jl
```diff
@@ -80,35 +67,41 @@ julia> W * W'
  0.0  1.0
 ```
 """
-whiten(a::AbstractMatrix, x::AbstractVecOrMat) = whiten!(similar(x), a, x)
-unwhiten(a::AbstractMatrix, x::AbstractVecOrMat) = unwhiten!(similar(x), a, x)
+whiten(a::AbstractMatrix{<:Real}, x::AbstractVecOrMat) = whiten(AbstractPDMat(a), x)
+unwhiten(a::AbstractMatrix{<:Real}, x::AbstractVecOrMat) = unwhiten(AbstractPDMat(a), x)
```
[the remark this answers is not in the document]
devmotion: Currently, `AbstractPDMat`s are `Real` matrices: https://github.com/JuliaStats/PDMats.jl/blob/9572e79761dfc1b0f7207f9b8bd6181764e0aadb/src/PDMats.jl#L36

andreasnoack (approved):

PR #184, opened 2023-09-26
Add FillArrays extension & bump Julia compat to >=1.6

devmotion:
This PR adds FillArrays as a weak dependency on Julia >= 1.9 and as a proper dependency on Julia < 1.9 (the latter might not be completely desirable, see https://github.com/JuliaStats/PDMats.jl/pull/182#issuecomment-1736061820, and alternatives would be to use Requires or reduce features on older Julia versions).

Currently, the only method in the extension is a definition of `AbstractPDMat` for `Diagonal` matrices with an `FillArray` as diagonal. This is motivated by the constructors of `MvNormal` in Distributions in https://github.com/JuliaStats/Distributions.jl/blob/87aebc29b2b9608801b70aae09fbc1d2dad56e3f/src/multivariate/mvnormal.jl#L201-L210: With the definition in this PR they could be simplified to
```julia
MvNormal(μ::AbstractVector{<:Real}, Σ::AbstractMatrix{<:Real}) = MvNormal(μ, AbstractPDMat(Σ))
MvNormal(μ::AbstractVector{<:Real}, Σ::UniformScaling{<:Real}) =
    MvNormal(μ, ScalMat(length(μ), Σ.λ))
```

This pattern is more generally useful, I imagine: Downstream packages that deal with pd matrices could be encouraged to use `AbstractPDMat(...)` for converting user-provided matrices to an optimized representation. The PR here adds such optimizations also for `Diagonal` with `FillArray`s as diagonal.

Edit: I dropped support for Julia < 1.6 in the PR. Tests were failing on Julia 1.0, even after adding compat with FillArrays 0.x - IIRC Pkg does not resolve again the package dependencies in the tests which caused issues due to incompatibilities between FillArrays and BandedMatrices.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -9,7 +9,7 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 SuiteSparse = "4607b0f0-06f3-5cda-b6b1-a6196a1729e9"
 
 [compat]
-FillArrays = "1"
+FillArrays = "0.12, 0.13, 1"
 julia = "1"
 
 [extensions]
--- a/Project.toml
+++ b/Project.toml
@@ -1,19 +1,28 @@
 name = "PDMats"
 uuid = "90014a1f-27ba-587c-ab20-58faa44d9150"
-version = "0.11.19"
+version = "0.11.20"
 
 [deps]
+FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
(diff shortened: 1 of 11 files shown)


Project.toml
```diff
@@ -1,19 +1,28 @@
...
 
 [deps]
+FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 SuiteSparse = "4607b0f0-06f3-5cda-b6b1-a6196a1729e9"
 
 [compat]
-julia = "1"
+FillArrays = "1"
```
devmotion, line 12: Alternatively, we could not support this feature on Julia < 1.9. But generally I don't like these inconsistencies.

devmotion:
Closed in favour of #192.

PR #187, opened 2023-10-02
Allow adddiag to work with abstract vectors

olivierverdier:


src/utils.jl
```diff
@@ -30,7 +30,7 @@ function _adddiag!(a::Union{Matrix, SparseMatrixCSC}, v::Real)
     return a
 end
 
-function _adddiag!(a::Union{Matrix, SparseMatrixCSC}, v::Vector, c::Real)
+function _adddiag!(a::Union{Matrix, SparseMatrixCSC}, v::AbstractVector, c::Real)
```
devmotion, line 33: Note that despite this change the functiob won't work for arbitrary `AbstractVector`s - the indices of `v` and the indices of `a` along each axis have to be identical.

devmotion (commented):
Can you add tests for the fix?

olivierverdier:
Superseded by #186.

Issue #22, opened 2015-04-13
Positivity check in PDiagMat and ScalMat constructors?


dmbates:
At present there is no check for positivity of the arguments that will be on the diagonal of a `PDiagMat` or `ScalMat`.  Probably a good idea to check that.
```
julia> using PDMats

julia> ScalMat(4,-1.5)
ScalMat(4,-1.5,-0.6666666666666666)

julia> PDiagMat([-1.,2.,-3.,4.])
PDiagMat(4,[-1.0,2.0,-3.0,4.0],[-1.0,0.5,-0.333333,0.25])

```

andreasnoack:
I think we need this. In particular now that we've also defined `isposdef` in https://github.com/JuliaStats/PDMats.jl/pull/123.

PR #188, opened 2023-10-04
Unify `PDMat` and `PDSparseMat` + move SparseArrays support to an extension

### LINKED ISSUE #174: Make SparseArrays a weak dependency?
From a very quick glance, it seems that a good portion of this package extends SparseArrays via `PDSparseMat` and related methods for it, but I wonder if, after splitting this from the main package, any dependency on SparseArrays.jl remains? If not, it would be very worthwhile to handle SparseArrays via the Pkg extension mechanism, to avoid the increased load time in Julia v1.10 due to SparseArrays.jl.

devmotion:
This PR unifies `PDMat` and `PDSparseMat` and makes `PDSparseMat` an (unexported) alias of `PDMat`. This makes it also possible to move support for `SparseArray`s to an extension, without any workarounds such as non-functional types in the main package (see #174).

Initially, `PDMat` and `PDSparseMat` only supported `Matrix{Float64}` and `SparseMatrixCSC{Float64}` matrices. This restriction was lifted in https://github.com/JuliaStats/PDMats.jl/pull/37 but the separate types with support for `Cholesky` and `CHOLMOD.Factor`s were kept.

~~Due to the refactoring of SuiteSparse and SparseArrays (it was integrated in SparseArrays in Julia 1.9), on Julia >= 1.9 only the SparseArrays dependency is needed and SuiteSparse only loads functionality from SparseArrays (https://github.com/JuliaSparse/SuiteSparse.jl/blob/e8285dd13a6d5b5cf52d8124793fc4d622d07554/src/SuiteSparse.jl). It seemed inconvenient to demand that users on Julia >= 1.9 load both SparseArrays and SuiteSparse for the sparse extension, and hence I decided to let the extension only depend on SparseArrays. IIUC, however, this means that we can't support Julia < 1.9 anymore since there SparseArrays does not contain the required `CHOLMOD` functionality. @dkarrasch, is this correct?~~

~~Since this means we drop support for the current LTS version, my suggestion would be to backport bugfixes to older Julia versions (at least >= 1.6) until another Julia version >= 1.9 becomes the new LTS version.~~

~~Edit: The PR keeps support for Julia 1.0-, I found a workaround: https://github.com/JuliaStats/PDMats.jl/pull/188#issuecomment-1746645655~~

Edit 2: The Julia version question above is moot by now — the package requires Julia >= 1.10, where `SparseArrays` always contains `CHOLMOD`. The extension therefore depends on `SparseArrays` alone, and the old compat workarounds have been dropped again.

## Breaking changes (0.12.0)

* `PDMat`'s third type parameter is widened from `C <: Cholesky{T}` to `F <: Factorization` and its second field is renamed `chol` -> `fact`, so that it can hold either a dense `Cholesky` or a sparse `CHOLMOD.Factor`. Use `cholesky(a)` rather than `a.chol`.
* `PDSparseMat` is no longer exported and no longer lives in the main package: it is a type alias defined inside the `SparseArrays` extension, so `PDMats.PDSparseMat` does not resolve. Sparse positive definite matrices are now built with `PDMat(::AbstractSparseMatrix)` (or `PDMat(::CHOLMOD.Factor)`) and are ordinary `PDMat`s.
* `inv` is no longer defined for sparse `PDMat`s. The previous `PDSparseMat` method could never run: `inv(::AbstractSparseMatrix)` deliberately errors in SparseArrays, because the inverse of a sparse matrix is generally dense. `inv` on a sparse `PDMat` now raises a `MethodError`, since the generic `inv(::Factorization)` requires `ldiv!(::CHOLMOD.Factor, ::Matrix)`, which SparseArrays does not define either -- the same gap that already keeps sparse matrices out of the `ldiv!` tests. Use `cholesky(a) \ I` (or densify) if you need an explicit inverse.
* `cor2cov`/`cov2cor` on a `PDMat` that is not backed by a `Cholesky` now agree with `cor2cov`/`cov2cor` on `Matrix(C)`. They previously fell through to the `AbstractPDMat` fallbacks, which compute `B * C * transpose(B)`, whereas `cor2cov`/`cov2cor` *overwrite* the diagonal (with `abs2.(s)` and `1` respectively) rather than scaling it. As a result `cov2cor` now returns an exactly unit diagonal, and input that violates the documented contract -- a `cor2cov` argument that is not a correlation matrix, or a `cov2cor` argument whose `s` are not its standard deviations -- throws a `PosDefException` instead of silently computing a different function.

## Other changes

* `X_A_Xt`/`Xt_A_X` with a `ScalMat` or `PDiagMat` second argument now return a positive definite type for *every* `PDMat`, not just `Cholesky`-backed ones. The methods in `src/congruence.jl` rebuild the Cholesky factor, so sparse `PDMat`s previously fell through to the generic methods and returned a `Symmetric`, losing the guarantee that file exists to provide. The new fallback rescales `mat` and factorizes the result, which works for any factorization and preserves sparsity.
* `src/` and `ext/` are at 100% line coverage. Beyond the lines this PR touched, that adds tests for `inv` (untested for every `AbstractPDMat` subtype -- `test_pdmat` only inverts the plain reference `Matrix`), for `PDMat{T, S}(::PDMat)`/`convert(::Type{PDMat{T, S}}, ::PDMat)` across differing element types, and for the `uplo === 'L'` branches of `cor2cov`/`cov2cor`, which were unreachable because `cholesky(::Matrix)` always returns `uplo == 'U'`.

Fixes #174.


test/testutils.jl
```diff
@@ -17,7 +23,7 @@ function test_pdmat(C, Cmat::Matrix;
                     t_cholesky::Bool=true,      # whether to test cholesky method
                     t_scale::Bool=true,         # whether to test scaling
                     t_add::Bool=true,           # whether to test pdadd
-		    t_det::Bool=true,           # whether to test det method
+		            t_det::Bool=true,           # whether to test det method
```
dkarrasch, line 26: This seems to be an accident.

ext/PDMatsSparseArraysExt/PDMatsSparseArraysExt.jl
```diff
@@ -0,0 +1,15 @@
+module PDMatsSparseArraysExt
+
+using PDMats
+using SparseArrays
+
+using PDMats.LinearAlgebra
```
dkarrasch, line 6: Out of curiosity, what exactly is the rationale here to not simply `using LinearAlgebra`?

dkarrasch (commented):
Thank you so much for taking up the challenge and moving this forward! I think this is a great service to the ecosystem. Perhaps we can make this work even for older Julia versions by `using SuiteSparse` or `SparseArrays.SuiteSparse` in the extension? On v1.9+, this should be a no-op, but on older versions (without Pkg extensions) this will just get loaded by including the extension code files. I'm not sure though if that would require "version-dependent" Project.toml files or other hassle.

test/testutils.jl
devmotion: I noticed the change in my git diff but in my editor the spacing actually looked correct with the change and wrong on the master branch. Weird, I'll fix it.

devmotion:
> Perhaps we can make this work even for older Julia versions by `using SuiteSparse` or `SparseArrays.SuiteSparse` in the extension? On v1.9+, this should be a no-op, but on older versions (without Pkg extensions) this will just get loaded by including the extension code files. I'm not sure though if that would require "version-dependent" Project.toml files or other hassle.

AFAICT on older Julia versions we would have to keep both SparseArrays and SuiteSparse as dependencies. However, if only Julia >= 1.9 the extension only depends on SparseArrays, then SuiteSparse would still be a regular dependency and since it depends on SparseArrays the extension would be loaded unconditionally. Alternatively, the extension could depend on both SparseArrays and SuiteSparse (or only SuiteSparse) - but then users would have to load both SparseArrays and SuiteSparse (or only SuiteSparse). Both alternatives seem suboptimal.

However, I just thought of another option that I will try later: Making SparseArrays and SuiteSparse both a weak dependency but letting the extension only depend on SparseArrays.

ext/PDMatsSparseArraysExt/PDMatsSparseArraysExt.jl
dkarrasch: Yeah, perhaps to avoid the impression that the extension introduces yet "additional" dependencies. Sounds reasonable.

test/testutils.jl
devmotion: Seems the file contained a few tabs. Should be fixed now.

ext/PDMatsSparseArraysExt/PDMatsSparseArraysExt.jl
devmotion: https://github.com/JuliaStats/LogExpFunctions.jl/pull/63

devmotion:
Seems to work 🙂 On Julia 1.0 and 1.6, PDMats depends on SparseArrays and SuiteSparse (https://github.com/JuliaStats/PDMats.jl/actions/runs/6404900221/job/17386347115?pr=188#step:5:36, https://github.com/JuliaStats/PDMats.jl/actions/runs/6404900221/job/17386347711?pr=188#step:5:27) and on Julia 1.9 and nightly neither SparseArrays nor SuiteSparse are dependencies (https://github.com/JuliaStats/PDMats.jl/actions/runs/6404900221/job/17386348265?pr=188#step:5:24, https://github.com/JuliaStats/PDMats.jl/actions/runs/6404900221/job/17386348761?pr=188#step:5:24) and the tests pass successfully with installing and loading only SparseArrays.

devmotion:
I'll update the PR once the other PRs (non-breaking bugfixes and possibly a new feature) are merged.

src/pdmat.jl
```diff
@@ -1,19 +1,19 @@
...
 Full positive definite matrix together with a Cholesky factorization object.
 """
-struct PDMat{T<:Real,S<:AbstractMatrix} <: AbstractPDMat{T}
+struct PDMat{T<:Real,S<:AbstractMatrix,C<:Factorization} <: AbstractPDMat{T}
     mat::S
-    chol::Cholesky{T,S}
+    chol::C
 
-    PDMat{T,S}(m::AbstractMatrix{T},c::Cholesky{T,S}) where {T,S} = new{T,S}(m,c)
+    PDMat{T,S,C}(m::AbstractMatrix{T}, c::Factorization) where {T,S,C} = new{T,S,C}(m,c)
```
timholy, line 8: Another case where the previous implementation was not ideal and we might as well fix it now: the `checksquare` and `size` checks on lines 12-15 should probably move here, into the inner constructor. Note that with the current implementation, it's possible to do this:
```julia
julia> chol = Cholesky(LowerTriangular([2 0; 1 2]))
Cholesky{Int64, Matrix{Int64}}
L factor:
2×2 LowerTriangular{Int64, Matrix{Int64}}:
 2  ⋅
 1  2

julia> PDMat{Int,Matrix{Int}}(rand(Int, 7, 2), chol)
7×7 PDMat{Int64, Matrix{Int64}}:
  6099129454893146728  -6401227969804411468  #undef  #undef  #undef  #undef  #undef
   320386356076645564   5388511647264862786  #undef  #undef  #undef  #undef  #undef
 -1586394843294801620   -539204836094727972  #undef  #undef  #undef  #undef  #undef
  5363274486172883276  -1644246604771485866  #undef  #undef  #undef  #undef  #undef
 -7714840428325231882   5088781302322332903  #undef  #undef  #undef  #undef  #undef
   371206992283242625    280269229271602249  #undef  #undef  #undef  #undef  #undef
 -4201226691849289285  -7373499862429006397  #undef  #undef  #undef  #undef  #undef
```
which is rather broken. It's much safer to put sanity checks into the inner constructor.

src/pdmat.jl
```diff
@@ -1,19 +1,19 @@
...
 end
 
-function PDMat(mat::AbstractMatrix,chol::Cholesky{T,S}) where {T,S}
+function PDMat(mat::AbstractMatrix{T}, chol::Cholesky) where {T<:Real}
     d = LinearAlgebra.checksquare(mat)
     if size(chol, 1) != d
         throw(DimensionMismatch("Dimensions of mat and chol are inconsistent."))
     end
```
timholy, line 15: These won't be needed if they move into the inner constructor.

src/pdmat.jl
```diff
@@ -1,19 +1,19 @@
 """
 Full positive definite matrix together with a Cholesky factorization object.
 """
-struct PDMat{T<:Real,S<:AbstractMatrix} <: AbstractPDMat{T}
+struct PDMat{T<:Real,S<:AbstractMatrix,C<:Factorization} <: AbstractPDMat{T}
     mat::S
-    chol::Cholesky{T,S}
+    chol::C
```
timholy, line 6: Not sure if this will be a breaking change, but if so should we change the fieldname to `fmat` or `fact` or something?

src/pdmat.jl
```diff
@@ -32,17 +32,16 @@ AbstractPDMat(A::Cholesky) = PDMat(A)
...
 end
 Base.convert(::Type{AbstractPDMat{T}}, a::PDMat) where {T<:Real} = convert(PDMat{T}, a)
 
 ### Basics
 
 Base.size(a::PDMat) = (a.dim, a.dim)
-Base.Matrix(a::PDMat) = copy(a.mat)
+Base.Matrix(a::PDMat) = Matrix(a.mat)
 LinearAlgebra.diag(a::PDMat) = diag(a.mat)
 LinearAlgebra.cholesky(a::PDMat) = a.chol
```
timholy, line 46:
```suggestion
LinearAlgebra.cholesky(a::PDMat{T,AbstractMatrix{T},<:Cholesky) where T = a.chol
```

timholy (commented):
Nice work!

One other thing we might want to look at, quite a few of the methods starting with `### (un)whitening` rely on the Choleksy factorization. If there isn't time to review/generalize those algorithms, perhaps it might make sense to define a `PDMatCholesky{T,S<:AbstractMatrix{T}} = PDMat{T,S,<:Cholesky}` and then restrict the argument of those methods to be a `PDMatCholesky`? That way we're at least not promising things we can't deliver.

src/pdmat.jl
devmotion: The current implementation was intended to provide a way to circumvent this check (since e.g. Zygote doesn't like them: https://github.com/JuliaStats/Distributions.jl/pull/1492) but in retrospect maybe this was not a good idea.

src/pdmat.jl
devmotion: This PR will be a breaking change anyway, so we could also rename the field. I guess I would prefer `fact`.

src/pdmat.jl
devmotion: I moved the checks to the inner constructor.

src/pdmat.jl
```diff
@@ -1,19 +1,19 @@
...
     mat::S
-    chol::Cholesky{T,S}
+    chol::C
 
-    PDMat{T,S}(m::AbstractMatrix{T},c::Cholesky{T,S}) where {T,S} = new{T,S}(m,c)
+    PDMat{T,S,C}(m::AbstractMatrix{T}, c::Factorization) where {T,S,C} = new{T,S,C}(m,c)
 end
 
-function PDMat(mat::AbstractMatrix,chol::Cholesky{T,S}) where {T,S}
+function PDMat(mat::AbstractMatrix{T}, chol::Cholesky) where {T<:Real}
```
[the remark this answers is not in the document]
devmotion: Maybe we should wait with generalizing the constructor until we actually support and test other factorizations.

timholy (approved):
Needs the conflicts resolved, but otherwise LGTM.

devmotion:
The PR is the result of some discussions I had with @andreasnoack about `PDSparseMat` and the SparseArrays dependency of PDMats, so I'd prefer to wait for his review.

timholy:
Right, I wasn't bumping you 🙂

README.md
```diff
@@ -27,16 +27,16 @@ Elemenent types are in princple all Real types, but in practice this is limited
 * `PDMat`: full covariance matrix, defined as
 
 ```julia
-struct PDMat{T<:Real,S<:AbstractMatrix} <: AbstractPDMat{T}
+struct PDMat{T<:Real,S<:AbstractMatrix{T},F<:Factorization} <: AbstractPDMat{T}
     mat::S                      # input matrix
-    chol::Cholesky{T,S}         # Cholesky factorization of mat
+    fact::F                     # factorization of mat
```
timholy, line 32: One needs an inner constructor that checks `isposdef(fact)` before calling `new`. Otherwise I could call
```julia
E = eigen(A)
PDMat{eltype(A),typeof(A),typeof(E)}(A, E)
```
even if `A` is not posdef.

test/pdmtypes.jl
```diff
@@ -296,4 +304,15 @@ using Test
...
+    # Ref https://github.com/JuliaStats/PDMats.jl/pull/207
+    # Note: `cholesky(::SymTridiagonal)` requires https://github.com/JuliaLang/julia/pull/44076
+    if VERSION >= v"1.8.0-DEV.1526"
+        @testset "PDMat from SymTridiagonal" begin
+            S = SymTridiagonal(fill(4, 4), fill(1, 3))
+            M = @inferred(PDMat(S))
+            @test M isa PDMat{Int,<:SymTridiagonal,<:Cholesky}
+            @test M == S
+        end
+    end
```
timholy, line 317: Probably want to add tests for what happens if one tries to construct with factorizations besides cholesky.

README.md
devmotion: Currently none of the other constructors enforces positive semi-definite matrices: https://github.com/JuliaStats/PDMats.jl/issues/22

So ideally this should be addressed more generally, to ensure that such checks are performed by all `AbstractPDMat` types. But I wonder if it should be left for a separate PR?

timholy: Sure. But it's not a hard change, we should just do it once other dust has settled.

dkarrasch:
Now that the lower compat has been raised to v1.10, this would be worthwhile to revisit and finish. I don't have current loading/latency times, but at the time this would have helped avoiding to load SparseArrays.jl unconditionally and therefore help all the dependent packages to keep loading times low(er).

PR #192, opened 2023-10-17
Add FillArrays extension

devmotion:
This is an alternative to #184 that keeps support for Julia 1.0. The issues with `Pkg.test` (incompatibilities of test dependencies with already installed package dependencies) can be avoided by enforcing the use of FillArrays 0.10 on Julia 1.0. Julia 1.0 is also supported by FillArrays 0.11 and 0.12 and hence I added them to the compat entry as well; however, the test dependencies BandedMatrices and ArrayLayouts can't be installed together with these more recent versions on Julia 1.0, hence I explicitly downgraded FillArrays to 0.10 in CI on Julia 1.0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,13 @@ jobs:
             ${{ runner.os }}-test-
             ${{ runner.os }}-
       - uses: julia-actions/julia-buildpkg@v1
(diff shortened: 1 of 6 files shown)


devmotion:
Closing since FillArrays would increase loading times significantly: https://github.com/JuliaStats/PDMats.jl/pull/182#issuecomment-1736061820

PR #203, opened 2024-01-30
Split the `whiten` and `unwhiten` docs.

timholy:


andreasnoack (approved):

timholy:
I don't have merge rights to JuliaStats, so someone will have to hit the button.

PR #209, opened 2024-10-02
Implement dot product

andreasKroepelin:
Currently, calling `LinearAlgebra.dot(x, A, y)` for an `AbstractPDMat` `A` falls back to the default implementation for `A::AbstractMatrix`. Especially for `A::PDiagMat` and `A::ScalMat` that is pretty wasteful, so I implemented the three-argument `dot` function in terms of `LinearAlgebra.Diagonal` and `LinearAlgebra.UniformScaling`, respectively.

Does it make sense to implement a special case for `A::PDMat` as well? Something similar to the `quad` function?
Also, is the test I added at a suitable location?

### DIFF
--- a/src/pdiagmat.jl
+++ b/src/pdiagmat.jl
@@ -203,3 +203,9 @@ function invquad(a::PDiagMat{<:Real,<:Vector}, x::Matrix)
     return invquad!(Vector{T}(undef, size(x, 2)), a, x)
 end
 
+### dot product
+
+function LinearAlgebra.dot(x::AbstractVector, a::PDiagMat, y::AbstractVector)
+    dot(x, Diagonal(a.diag), y)
+end
+
--- a/src/scalmat.jl
+++ b/src/scalmat.jl
@@ -187,3 +187,9 @@ function Xt_invA_X(a::ScalMat, x::Matrix{<:Real})
     @check_argdims LinearAlgebra.checksquare(a) == size(x, 1)
     return Symmetric(_rdiv!(transpose(x) * x, a.value))
 end
+
+### dot product
+
+function LinearAlgebra.dot(x::AbstractVector, a::ScalMat, y::AbstractVector)
+    dot(x, UniformScaling(a.value), y)
+end
--- a/test/generics.jl
+++ b/test/generics.jl
@@ -28,3 +28,13 @@ end
 @test isposdef(PDMat([1.0 0.0; 0.0 1.0]))
 @test isposdef(PDiagMat([1.0, 1.0]))
 @test isposdef(ScalMat(2, 3.0))
+
+@testset "dot products" begin
+    pm1 = PDiagMat([1., 2., 3.])
+    pm2 = ScalMat(3, 2.)
+    x = [13., 42., .7]
+    y = [0.5, .125, 20.24]
+
+    @test dot(x, pm1, y) ≈ dot(x, Matrix(pm1), y)
+    @test dot(x, pm2, y) ≈ dot(x, Matrix(pm2), y)
+end


andreasKroepelin:
I'm puzzled why the test is failing, it works for me locally...

src/pdiagmat.jl
```diff
@@ -203,3 +203,9 @@ function invquad(a::PDiagMat{<:Real,<:Vector}, x::Matrix)
     return invquad!(Vector{T}(undef, size(x, 2)), a, x)
 end
 
+### dot product
+
+function LinearAlgebra.dot(x::AbstractVector, a::PDiagMat, y::AbstractVector)
```
devmotion, line 208: Should we restrict this to
```suggestion
function LinearAlgebra.dot(x::AbstractVector{<:Real}, a::PDiagMat, y::AbstractVector{<:Real})
```
?

src/pdiagmat.jl
```diff
@@ -203,3 +203,9 @@ function invquad(a::PDiagMat{<:Real,<:Vector}, x::Matrix)
     return invquad!(Vector{T}(undef, size(x, 2)), a, x)
 end
 
+### dot product
+
```
devmotion, line 207: This has to be put behind a version check:
```suggestion
# https://github.com/JuliaLang/julia/commit/2425ae760fb5151c5c7dd0554e87c5fc9e24de73
if VERSION >= v"1.4.0-DEV.92"
```

src/pdiagmat.jl
```diff
@@ -203,3 +203,9 @@ function invquad(a::PDiagMat{<:Real,<:Vector}, x::Matrix)
...
+function LinearAlgebra.dot(x::AbstractVector, a::PDiagMat, y::AbstractVector)
+    dot(x, Diagonal(a.diag), y)
+end
+
```
devmotion, line 211:
```suggestion
end
```

src/scalmat.jl
```diff
@@ -187,3 +187,9 @@ function Xt_invA_X(a::ScalMat, x::Matrix{<:Real})
     @check_argdims LinearAlgebra.checksquare(a) == size(x, 1)
     return Symmetric(_rdiv!(transpose(x) * x, a.value))
 end
+
+### dot product
+
```
devmotion, line 192: Same here:
```suggestion

# https://github.com/JuliaLang/julia/commit/2425ae760fb5151c5c7dd0554e87c5fc9e24de73
if VERSION >= v"1.4.0-DEV.92"
```

src/scalmat.jl
```diff
@@ -187,3 +187,9 @@ function Xt_invA_X(a::ScalMat, x::Matrix{<:Real})
...
+function LinearAlgebra.dot(x::AbstractVector, a::ScalMat, y::AbstractVector)
```
devmotion, line 193: We should check that the dimensions match:
```suggestion
function LinearAlgebra.dot(x::AbstractVector, a::ScalMat, y::AbstractVector)
    @check_argdims LinearAlgebra.checksquare(a) == length(x) == length(y)
```

src/scalmat.jl
```diff
@@ -187,3 +187,9 @@ function Xt_invA_X(a::ScalMat, x::Matrix{<:Real})
...
+    dot(x, UniformScaling(a.value), y)
```
devmotion, line 194: I guess even simpler would be
```suggestion
    a.value * dot(x, y)
```
?

src/scalmat.jl
```diff
@@ -187,3 +187,9 @@ function Xt_invA_X(a::ScalMat, x::Matrix{<:Real})
...
+end
```
devmotion, line 195:
```suggestion
end
end
```

test/generics.jl
```diff
@@ -28,3 +28,13 @@ end
 @test isposdef(PDMat([1.0 0.0; 0.0 1.0]))
 @test isposdef(PDiagMat([1.0, 1.0]))
 @test isposdef(ScalMat(2, 3.0))
+
```
devmotion, line 31:
```suggestion

# https://github.com/JuliaLang/julia/commit/2425ae760fb5151c5c7dd0554e87c5fc9e24de73
if VERSION >= v"1.4.0-DEV.92"
```

test/generics.jl
```diff
@@ -28,3 +28,13 @@ end
...
+@testset "dot products" begin
+    pm1 = PDiagMat([1., 2., 3.])
+    pm2 = ScalMat(3, 2.)
+    x = [13., 42., .7]
+    y = [0.5, .125, 20.24]
+
+    @test dot(x, pm1, y) ≈ dot(x, Matrix(pm1), y)
+    @test dot(x, pm2, y) ≈ dot(x, Matrix(pm2), y)
+end
```
devmotion, line 40:
```suggestion
end
end
```

PR #212, opened 2024-12-10
More conversion in `PDMat`

timholy:
This allows one to convert both the eltype and matrix type. In practice, my main use for this is to  support generic programming with both `Matrix` and `StaticArrays.SMatrix`.

Perhaps the only potentially-controversial aspect is moving the dimensionality check into the inner constructor. If we want to make it impossible to construct an inconsistent `PDMat`, that's where it has to be.

### DIFF
--- a/src/chol.jl
+++ b/src/chol.jl
@@ -73,7 +73,7 @@ function invquad(A::Cholesky, x::AbstractVector)
     @check_argdims size(A, 1) == size(x, 1)
     return sum(abs2, chol_lower(A) \ x)
 end
-function invquad(A::Cholesky, X::AbstractMatrix) 
+function invquad(A::Cholesky, X::AbstractMatrix)
     @check_argdims size(A, 1) == size(X, 1)
     Z = chol_lower(A) \ X
     return vec(sum(abs2, Z; dims=1))
--- a/src/pdmat.jl
+++ b/src/pdmat.jl
@@ -1,23 +1,54 @@
 """
 Full positive definite matrix together with a Cholesky factorization object.
 """
-struct PDMat{T<:Real,S<:AbstractMatrix} <: AbstractPDMat{T}
+struct PDMat{T<:Real,S<:AbstractMatrix{T}} <: AbstractPDMat{T}
     mat::S
     chol::Cholesky{T,S}
 
-    PDMat{T,S}(m::AbstractMatrix{T},c::Cholesky{T,S}) where {T,S} = new{T,S}(m,c)
-end
-
-function PDMat(mat::AbstractMatrix,chol::Cholesky{T,S}) where {T,S}
-    d = LinearAlgebra.checksquare(mat)
-    if size(chol, 1) != d
-        throw(DimensionMismatch("Dimensions of mat and chol are inconsistent."))
+    function PDMat{T,S}(m::AbstractMatrix, c::Cholesky) where {T<:Real,S<:AbstractMatrix{T}}
+        d = LinearAlgebra.checksquare(m)
+        if size(c, 1) != d
+            throw(DimensionMismatch("Dimensions of mat and chol are inconsistent."))
+        end
+        # in principle we might want to check that `c` is a Cholesky factorization of `m`,
+        # but that's slow
+        return new{T,S}(m,c)
     end
-    PDMat{T,S}(convert(S, mat), chol)
 end
+function PDMat{T}(m::AbstractMatrix, c::Cholesky) where T<:Real
+    c = convert(Cholesky{T}, c)
+    return PDMat{T,typeof(c.factors)}(m, c)
+end
+PDMat(mat::AbstractMatrix,chol::Cholesky{T,S}) where {T<:Real,S<:AbstractMatrix{T}} = PDMat{T,S}(mat, chol)
 
+# Construction from another PDMat
+PDMat{T,S}(pdm::PDMat{T,S}) where {T<:Real,S<:AbstractMatrix{T}} = pdm  # since PDMat doesn't support `setindex!` it's not mutable (xref https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Mutable-collections)
+PDMat{T,S}(pdm::PDMat) where {T<:Real,S<:AbstractMatrix{T}} = PDMat{T,S}(pdm.mat, pdm.chol)
+PDMat{T}(pdm::PDMat{T}) where T<:Real = pdm
+PDMat{T}(pdm::PDMat) where T<:Real = PDMat{T}(pdm.mat, pdm.chol)
+PDMat(pdm::PDMat) = pdm
+
+# Construction from an AbstractMatrix
+function PDMat{T,S}(mat::AbstractMatrix) where {T<:Real,S<:AbstractMatrix{T}}
+    mat = convert(S, mat)
+    return PDMat{T,S}(mat, cholesky(mat))
+end
+function PDMat{T}(mat::AbstractMatrix) where T<:Real
+    mat = convert(AbstractMatrix{T}, mat)
+    return PDMat{T}(mat, cholesky(mat))
+end
 PDMat(mat::AbstractMatrix) = PDMat(mat, cholesky(mat))
-PDMat(fac::Cholesky) = PDMat(AbstractMatrix(fac), fac)
+
+# Construction from a Cholesky factorization
+function PDMat{T,S}(c::Cholesky) where {T<:Real,S<:AbstractMatrix{T}}
+    c = convert(Cholesky{T,S}, c)
+    return PDMat{T,S}(AbstractMatrix(c), c)
+end
+function PDMat{T}(c::Cholesky) where T<:Real
+    c = convert(Cholesky{T}, c)
+    return PDMat{T}(AbstractMatrix(c), c)
+end
+PDMat(c::Cholesky) = PDMat(AbstractMatrix(c), c)
 
 function Base.getproperty(a::PDMat, s::Symbol)
     if s === :dim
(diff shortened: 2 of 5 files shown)


devmotion:
Did you check whether this is covered by https://github.com/JuliaStats/PDMats.jl/pull/188? At least it seems the matrix element type is restricted in the same way in #188.

timholy:
> Did you check whether this is covered by https://github.com/JuliaStats/PDMats.jl/pull/188? At least it seems the matrix element type is restricted in the same way in https://github.com/JuliaStats/PDMats.jl/pull/188.

The eltype restriction is just a small cleanup. The key aspect of this is the "constructor cascade" that supports conversion/coercion via `PDMat{T}(args...)` or `PDMat{T,S}(args...)`. This is needed to support, e.g., `push!(pdmats, pdmat)` where `eltype(pdmats)` may not be identical to `typeof(pdmat)`.

Also, #188 is a pretty deep change, and it's been open for more than a year. I hoped this could get merged quickly and a new release made.

timholy:
Concretely, if one runs the test block "test the functionality" from this PR on `master`, one gets this (with `T = Float32`):
```julia
Testing PDMat{Float32, Matrix{Float32}} of size (3, 3)
PDMat from Matrix: Error During Test at REPL[8]:7
  Got exception outside of a @test
  MethodError: no method matching (PDMat{Float64})(::Matrix{Float32})
  Stacktrace:
```
and several similar errors.

src/pdmat.jl
```diff
@@ -1,23 +1,49 @@
...
+        d = LinearAlgebra.checksquare(m)
+        if size(c, 1) != d
+            throw(DimensionMismatch("Dimensions of mat and chol are inconsistent."))
+        end
+        # in principle we might want to check that `c` is a Cholesky factorization of `m`,
+        # but that's slow
+        return new{T,S}(m,c)
+    end
 end
+PDMat{T,S}(pdm::PDMat) where {T,S} = PDMat{T,S}(pdm.mat, pdm.chol)
```
devmotion, line 18: Is it common to define these constructors with type parameters? Or even necessary? It seems it's not done for any of the other matrix types in PDMats. I assumed that defining the relevant `convert` methods and the optimization
```suggestion
PDMat(pdm::PDMat) = pdm
```
would be sufficient?

timholy: To be clear, this is one of the elegances of Julia's type system. `Matrix{T}` is a concrete type, and you should be able to construct the concrete type of your choosing: you shouldn't be doomed to just use `Matrix(A)`, because that does not allow you to control the eltype. `Cholesky{T}` is not a concrete type, because Cholesky depends on two type parameters. A very cool aspect of Julia's type system is that you can specify the first concretely and allow the second to be computed from the inputs, *or* you can specify both concretely. Thus you want methods `Cholesky{T,S})(args...)`, `Cholesky{T}(args...)` and `Cholesky(args...)` (all 3 variants).

That's what this PR adds for `PDMat`. More of Julia's linear algebra ecosystem should follow suit, but you have to start somewhere.

timholy:
If you're OK with it, I'd plan to solve the failure on Julia 1.0 by dropping support for anything before 1.10. Since this package is still in 0.x releases, we don't have to bump the minor version to do so.

src/pdmat.jl
devmotion: I see. I was not sure regarding the conventions for `AbstractMatrix` constructors and conversions and was wondering whether we could rely on some existing functionality in LinearAlgebra. Based on the linked PR, it seems the approach in this PR (defining constructors instead of `convert`) follows the convention in LinearAlgebra.

src/pdmat.jl
```diff
@@ -1,23 +1,49 @@
...
 
-function PDMat(mat::AbstractMatrix,chol::Cholesky{T,S}) where {T,S}
-    d = LinearAlgebra.checksquare(mat)
-    if size(chol, 1) != d
-        throw(DimensionMismatch("Dimensions of mat and chol are inconsistent."))
-    end
-    PDMat{T,S}(convert(S, mat), chol)
+function PDMat{T}(m::AbstractMatrix, c::Cholesky) where T
+    c = Cholesky{T}(c)
```
devmotion, line 21: This seems unnecessary if `c isa Cholesky{T}` - is it guaranteed to be a no-op in that case? Or is it necessary to use `convert` as for `m` in the other constructors?
```suggestion
    c = convert(Cholesky{T}, c)
```

src/pdmat.jl
```diff
@@ -1,23 +1,49 @@
...
+    return PDMat{T,mattype(c)}(m, c)
 end
+PDMat{T}(pdm::PDMat) where T = PDMat{T}(pdm.mat, pdm.chol)
```
devmotion, line 24: Maybe group this together with the definition of `PDMat{T,S}(pdm::PDMat)` above? The large number of constructors makes it a bit confusing to read.

src/pdmat.jl
```diff
@@ -1,23 +1,49 @@
 """
 Full positive definite matrix together with a Cholesky factorization object.
 """
-struct PDMat{T<:Real,S<:AbstractMatrix} <: AbstractPDMat{T}
+struct PDMat{T<:Real,S<:AbstractMatrix{T}} <: AbstractPDMat{T}
     mat::S
     chol::Cholesky{T,S}
 
-    PDMat{T,S}(m::AbstractMatrix{T},c::Cholesky{T,S}) where {T,S} = new{T,S}(m,c)
+    function PDMat{T,S}(m::AbstractMatrix, c::Cholesky) where {T,S}
```
devmotion, line 8: Shouldn't this ensure
```suggestion
    function PDMat{T,S}(m::AbstractMatrix{T}, c::Cholesky{T,S}) where {T,S}
```
as on the master branch?

src/pdmat.jl
```diff
@@ -1,23 +1,49 @@
...
+end
+function PDMat{T}(mat::AbstractMatrix) where T
+    mat = convert(AbstractMatrix{T}, mat)
+    return PDMat{T}(mat, cholesky(mat))
+end
 PDMat(mat::AbstractMatrix) = PDMat(mat, cholesky(mat))
-PDMat(fac::Cholesky) = PDMat(AbstractMatrix(fac), fac)
+
+function PDMat{T,S}(c::Cholesky) where {T,S}
+    c = Cholesky{T,S}(c)
```
devmotion, line 39: Similarly, is this guaranteed to be a no-op if `c isa Cholesky{T,S}`? Or should we use
```suggestion
    c = convert(Cholesky{T,S}, c)
```
?

src/pdmat.jl
```diff
@@ -1,23 +1,49 @@
...
+    return PDMat{T,S}(AbstractMatrix(c), c)
+end
+function PDMat{T}(c::Cholesky) where T
+    c = Cholesky{T}(c)
```
devmotion, line 43: Same here, is there any difference between using the constructor and
```suggestion
    c = convert(Cholesky{T}, c)
```
?

src/pdmat.jl
```diff
@@ -30,13 +56,11 @@ Base.propertynames(::PDMat) = (:mat, :chol, :dim)
 AbstractPDMat(A::Cholesky) = PDMat(A)
 
 ### Conversion
+Base.convert(::Type{PDMat{T,S}}, a::PDMat{T,S}) where {T<:Real,S<:AbstractMatrix{T}} = a
```
devmotion, line 59: Is the convention to define this no-op `convert` method? Or should one rather define
```julia
PDMat{T,S}(pdm::PDMat{T,S}) where {T,S} = pdm
```
?

src/chol.jl
```diff
@@ -25,6 +25,8 @@ if HAVE_CHOLMOD
     chol_upper(cf::CholTypeSparse) = cf.UP
 end
 
+mattype(::Cholesky{T,S}) where {T,S} = S
```
devmotion, line 28: Maybe omit it since it's only used once?

src/pdmat.jl
timholy: No. This variant allows you to construct the chosen type regardless of the types of the inputs (as long as they are `AbstractMatrix` and `Cholesky`). Anything that doesn't match will be `convert`ed.

src/chol.jl
timholy: Sure, but we need a replacement. We need to extract the `S` from `Cholesky{T,S}` but note the `S` may not be the same `S` as for the input to the function that uses `mattype`. (The `convert(Cholesky{T}, c)` might change the `S` type.)

src/pdmat.jl
timholy: I like your suggestion, implemented.

src/pdmat.jl
timholy: With the right constructors it's OK, but until you pointed this out we were lacking them. I've now deleted several `convert` methods.

src/chol.jl
devmotion: It seems on the master branch we use `typeof(c.factors)`.

src/pdmat.jl
```diff
@@ -31,12 +62,8 @@ AbstractPDMat(A::Cholesky) = PDMat(A)
 
 ### Conversion
 Base.convert(::Type{PDMat{T}}, a::PDMat{T}) where {T<:Real} = a
```
devmotion, line 64: It seems this can be removed?
```suggestion
```

src/pdmat.jl
```diff
@@ -1,23 +1,54 @@
...
-    PDMat{T,S}(convert(S, mat), chol)
 end
+function PDMat{T}(m::AbstractMatrix, c::Cholesky) where T
+    c = convert(Cholesky{T}, c)
+    return PDMat{T,mattype(c)}(m, c)
+end
+PDMat(mat::AbstractMatrix,chol::Cholesky{T,S}) where {T,S} = PDMat{T,S}(mat, chol)
+
+# Construction from another PDMat
+PDMat{T,S}(pdm::PDMat{T,S}) where {T,S} = pdm
```
devmotion, line 25: I wonder if the constructor should return an actual copy given https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Mutable-collections. On the other hand, based on the PR you linked to LinearAlgebra doesn't seem to follow this convention?

timholy: It's technically not a mutable collection: `pdm.mat = rand(3,3)` will fail.

devmotion:
> If you're OK with it, I'd plan to solve the failure on Julia 1.0 by dropping support for anything before 1.10.

I'm fine with that but I'm still curious which and why some of the changes are problematic on Julia 1.0.

src/pdmat.jl
timholy: Not if you want it to remain a no-op. The generic fallback `convert(::Type{T}, x::T)` defined in `Core.Compiler` is of lower specificity than the method below this one.

timholy: It seems we have to keep the method below it. And then this one is necessary if we want it to remain a no-op; the default fallback `convert(::Type{T}, x::T) where T = x` defined in `Core.Compiler` is too low specificity to be the dispatch target, given the method below this one.

timholy:
> I'm fine with that but I'm still curious which and why some of the changes are problematic on Julia 1.0.

I'm a little puzzled about that as well. But it's only the `@allocated` test that fails and that is indeed pretty sensitive (especially in old Julias) to minor details. Looking at the produced code
```julia
julia> @code_typed PDMat(S, C)
CodeInfo(
22 1 ─ %1 = %new(PDMat{Float64,SArray{Tuple{4,4},Float64,2,16}}, mat, chol)::PDMat{Float64,SArray{Tuple{4,4},Float64,2,16}}                       │╻ Type
   └──      return %1                                                                                                                             │
) => PDMat{Float64,SArray{Tuple{4,4},Float64,2,16}}
```
it's a little weird that it allocates at all.

I dunno, I am not personally that interested in pursuing why Julia 1.0 allocates in this circumstance.

timholy:
This should be ready. Let me know if you want more changes. And thanks for the review, it was very helpful.

devmotion (approved):

PR #213, opened 2024-12-11
Update Julia compat to 1.10 and update CI setup

devmotion:
In https://github.com/JuliaStats/PDMats.jl/pull/212, it was proposed to update the Julia compat to 1.10. I opened a separate PR for this move since there are quite many `VERSION` checks that we could (and should) get rid of in this update but these changes are really unrelated to #212 otherwise.

I used the opportunity to update the CI setup, and in particular add tests with `min`, `lts` and `pre`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
   pull_request:
 jobs:
   test:
-    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ (matrix.arch == '') && runner.arch || matrix.arch }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -6,50 +6,44 @@ on:
...
-          - '1.0'
-          - '1.6'
+          - 'min'
+          - 'lts'
           - '1'
-          - nightly
+          - 'pre'
         os:
-          - ubuntu-latest
+          - ubuntu-latest # Apple Silicon
```
timholy, line 20: Did you mean to have this comment on this line or the following one?

timholy (approved):

.github/workflows/ci.yml
devmotion: Ah good catch, of course it should be moved to the following line.

.github/workflows/ci.yml
```diff
@@ -6,50 +6,44 @@ on:
...
+          - macOS-latest
```
devmotion, line 21:
```suggestion
          - ubuntu-latest
          - macOS-latest # Apple Silicon
```

timholy:
I wonder if it makes sense to bump the patch version here too, get a release out, and then bump again with #212?

timholy:
Agreed. There's a couple of minor changes that haven't been released, though: https://github.com/JuliaStats/PDMats.jl/compare/v0.11.31...master

Probably doesn't matter, though.

timholy:
Bump. I think this is pretty safe, so no need to wait too long for @andreasnoack's review.

PR #219, opened 2025-04-18
Add JuliaFormatter

devmotion:
Same as https://github.com/JuliaStats/GLM.jl/pull/603. I used YAS style to ensure a consistent style throughout JuliaStats.

Edit: Marking as a draft until the discussion in https://github.com/JuliaStats/StatsFuns.jl/pull/181 is resolved.

PR #220, opened 2025-04-23
Format with Runic

devmotion:
Alternative to #219. Same as https://github.com/JuliaStats/StatsFuns.jl/pull/182.

### DIFF
--- a/.github/workflows/Format.yml
+++ b/.github/workflows/Format.yml
@@ -1,0 +1,22 @@
+name: Check formatting
+on:
+  push:
+    branches:
+      - 'master'
+      - 'release-'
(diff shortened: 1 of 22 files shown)


andreasnoack:
It looks like the extra allocations on Mac could be triggered by the new Julia patch release. Seems a bit odd that it only happens on Mac. Probably not much we can do here but adjust the tests.

devmotion:
I adjusted the tests: https://github.com/JuliaStats/PDMats.jl/pull/220/commits/29a15948af4273c4b781c57f7131c2e13a84342d

andreasnoack (approved):

PR #222, opened 2025-05-09
Overload congruent transforms for AbstractPDMat return types

### LINKED ISSUE #132: Constructor for PDMat with Cholesky pre- and post- multiplied by diagonal matrix
Suppose a user has a `Cholesky` decomposition of a matrix `R` and a vector of positive entries `σ` and wishes to create `Σ = PDMat(cholesky(σ .* Matrix(R) .* σ'))`. This is common in Bayesian inference. e.g. a Cholesky factor of a correlation matrix is drawn from a `LKJCholesky` distribution, while a vector of standard deviations is drawn from some other distribution, and one wants to construct the Cholesky factorization of the covariance matrix for use in `MvNormal`.

This can be done much more efficiently with `Σ = PDMat(Cholesky(R.uplo == 'U' ? R.U * Diagonal(σ) : Diagonal(σ) * R.L, R.uplo, 0))`. But this is still quite long. It would be ideal to have either a constructor for `PDMat` or some other utility function like the following:

```julia
function PDMat(R::Cholesky, σ::AbstractVector)
    return PDMat(Cholesky(R.uplo == 'U' ? R.U * Diagonal(σ) : Diagonal(σ) * R.L, R.uplo, 0))
end
function PDMat(R::Cholesky, σ::UniformScaling)
    return PDMat(Cholesky(R.factors * σ, R.uplo, 0))
end
PDMat(R::Cholesky, σ::Number) = PDMat(R, σ * I)
```

See also discussions in https://github.com/JuliaStats/Distributions.jl/issues/1336 and https://github.com/TuringLang/Turing.jl/issues/1629#issuecomment-851959832.

sethaxen:
As discussed in https://github.com/JuliaStats/PDMats.jl/issues/132#issuecomment-2830828344, this PR overloads `X_A_Xt`, `Xt_A_X`, `X_invA_Xt`, and `Xt_invA_X` to return the appropriate PDMats type when we can guarantee the return type is an `AbstractPDMat` and when it doesn't require computing a new composition. Fixes #132.

### DIFF
--- a/src/PDMats.jl
+++ b/src/PDMats.jl
@@ -53,6 +53,7 @@ include("scalmat.jl")
 
 include("generics.jl")
 include("addition.jl")
+include("congruent.jl")
 
 include("deprecates.jl")
 
--- a/src/congruent.jl
+++ b/src/congruent.jl
@@ -1,0 +1,53 @@
+# Congruent transforms X_A_Xt, Xt_A_X, X_invA_Xt, Xt_invA_X for guaranteed PD return values
+
+for f in (:X_A_Xt, :Xt_A_X)
+    @eval begin
+        function $(f)(A::ScalMat, B::ScalMat)
+            @check_argdims A.dim == B.dim
+            return ScalMat(A.dim, abs2(B.value) * A.value)
+        end
+        $(f)(A::PDiagMat, B::PDiagMat) = PDiagMat(abs2.(B.diag) .* A.diag)
+        $(f)(A::PDiagMat, B::ScalMat) = PDiagMat(abs2(B.value) .* A.diag)
+        function $(f)(A::ScalMat, B::PDiagMat)
+            @check_argdims A.dim == size(B, 1)
+            return PDiagMat(abs2.(B.diag) .* A.value)
+        end
+        function $(f)(A::PDMat, B::ScalMat)
+            @check_argdims B.dim == size(A, 1)
+            b2 = abs2(B.value)
+            mat = b2 * A.mat
+            uplo = A.chol.uplo
+            if uplo === 'U'
+                factors = A.chol.factors * B.value'
+            else
+                factors = B.value * A.chol.factors
+            end
+            chol = Cholesky(factors, uplo, A.chol.info)
+            return PDMat(mat, chol)
+        end
+        function $(f)(A::PDMat, B::PDiagMat)
+            b = B.diag
+            mat = A.mat .* (b .* b')
+            uplo = A.chol.uplo
+            if uplo === 'U'
+                factors = A.chol.factors .* b'
+            else
+                factors = b .* A.chol.factors
+            end
+            chol = Cholesky(factors, uplo, A.chol.info)
+            return PDMat(mat, chol)
+        end
+    end
+end
+
+for f in (:X_invA_Xt, :Xt_invA_X)
+    @eval begin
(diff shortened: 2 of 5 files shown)


src/congruence.jl
```diff
@@ -0,0 +1,53 @@
+# Congruent transforms X_A_Xt, Xt_A_X, X_invA_Xt, Xt_invA_X for guaranteed PD return values
+
+for f in (:X_A_Xt, :Xt_A_X)
+    @eval begin
+        function $(f)(A::ScalMat, B::ScalMat)
+            @check_argdims A.dim == B.dim
+            return ScalMat(A.dim, abs2(B.value) * A.value)
+        end
+        $(f)(A::PDiagMat, B::PDiagMat) = PDiagMat(abs2.(B.diag) .* A.diag)
+        $(f)(A::PDiagMat, B::ScalMat) = PDiagMat(abs2(B.value) .* A.diag)
```
devmotion, line 10: AFAICT we also have to check that dimensions match in this function:
```suggestion
        function $(f)(A::PDiagMat, B::ScalMat)
            @check_argdims size(A, 1) == B.dim
            return PDiagMat(abs2(B.value) .* A.diag)
        end
```

src/congruence.jl
```diff
@@ -0,0 +1,53 @@
...
+            @check_argdims B.dim == size(A, 1)
+            b2 = abs2(B.value)
+            mat = b2 * A.mat
+            uplo = A.chol.uplo
+            if uplo === 'U'
+                factors = A.chol.factors * B.value'
+            else
+                factors = B.value * A.chol.factors
+            end
+            chol = Cholesky(factors, uplo, A.chol.info)
```
devmotion, line 25: Since `ScalMat` is a `AbstractMatrix{<:Real}`, it seems we could merge the two branches:
```suggestion
            factors = B.value * A.chol.factors
            chol = Cholesky(factors, A.chol.uplo, A.chol.info)
```

src/congruence.jl
```diff
@@ -0,0 +1,53 @@
...
+            uplo = A.chol.uplo
+            if uplo === 'U'
+                factors = A.chol.factors .* b'
+            else
+                factors = b .* A.chol.factors
+            end
+            chol = Cholesky(factors, uplo, A.chol.info)
+            return PDMat(mat, chol)
+        end
+    end
```
devmotion, line 40: For completeness, probably we should also add definitions for first arguments of types `PDSparseMat` and `Cholesky`.

src/congruence.jl
```diff
@@ -0,0 +1,53 @@
...
+for f in (:X_invA_Xt, :Xt_invA_X)
+    @eval begin
+        $(f)(A::ScalMat, B::ScalMat) = B * (A \ B)
+        $(f)(A::PDiagMat, B::PDiagMat) = PDiagMat(B.diag .* (A.diag .\ B.diag))
+        $(f)(A::PDiagMat, B::ScalMat) = PDiagMat(B.value .* (A.diag .\ B.value))
+        function $(f)(A::ScalMat, B::PDiagMat)
+            @check_argdims A.dim == size(B, 1)
+            return PDiagMat(B.diag .* (A.value .\ B.diag))
+        end
+    end
```
devmotion, line 52: What's the reason to omit definitions for `PDMat` (and `PDSparseMat` and `Cholesky`)? That they wouldn't be more efficient than the generic definitions?

PR #236, opened 2026-06-24
Fix a few bugs

devmotion:
I noticed these (unrelated) bugs while updating #188:

- The `quad`/`invquad` specializations for `PDMat` were dispatched on `PDMat{<:Real, <:Vector}`. But the second type parameter of `PDMat` is the type of the wrapped matrix (`<:AbstractMatrix`), so this signature can never be satisfied and the specializations were never used — instead the generic fallbacks, which materialize `chol_upper(a) * x`, were called. Changing it to `<:Matrix` restores the intended implementation: for an `n×n` matrix and an `n×k` input the specialization reuses a single length-`n` buffer, so it allocates `O(n + k)` instead of the `O(n*k)` temporary of the fallback.
- `\` and `/` for `PDSparseMat` unnecessarily required the element types of the matrix and the right-hand side to match. Since CHOLMOD only supports `Float64`/`ComplexF64` anyway, this excluded e.g. integer or `Float32` inputs that the other types accept. I relaxed the signatures to `AbstractVecOrMat{<:Real}` and promote the element type of the result.
- The right-division method `/(::AbstractVecOrMat, ::PDSparseMat)` was in fact dead: CHOLMOD provides no right division, so `x / a.chol` always threw a `MethodError`, and every test skipped the sparse right-division path (so it was also uncovered). Since `a` is symmetric, `x / a == (a \ xᵀ)ᵀ`, so I route through the (working, covered) left-division method and enable the previously-skipped right-division tests for `PDSparseMat`.
- I removed the self-recursive definition `*(a::AbstractPDMat, c::Real) = a * c` in `generics.jl`. It is shadowed by the concrete methods of the built-in types but leads to a `StackOverflowError` for other `AbstractPDMat` subtypes.
- For consistency with the other types I defined `Matrix{T}(::ScalMat)` instead of `Matrix(::ScalMat)` (the untyped constructor is provided by `Base`).

Regression tests are added in `test/typeparams.jl`.

The allocation test in `test/chol.jl` was marked broken on Apple since Julia 1.11.5. The bound turns out to be unreliable there in a machine-dependent way: it is met on my Apple silicon machine (Julia 1.12.6) but exceeded on the CI Apple runner for the same version, so neither value of the `broken` marker works for both (it errors as "unexpectedly passing" where it passes, and fails where it doesn't). I therefore `skip` the allocation check on Apple (Julia >= 1.11.5) instead. The marker only ever applied to Apple, so other CI configurations are unaffected.

### DIFF
--- a/src/pdsparsemat.jl
+++ b/src/pdsparsemat.jl
@@ -76,8 +76,9 @@ function \(a::PDSparseMat, x::AbstractVecOrMat{<:Real})
 end
 function /(x::AbstractVecOrMat{<:Real}, a::PDSparseMat)
     @check_argdims a.dim == size(x, 2)
-    T = promote_type(eltype(a), eltype(x))
-    return convert(Array{T}, convert(Array{Float64}, x) / a.chol)
+    # CHOLMOD does not support `/`, but `a` is symmetric: `x / a == (a \ xᵀ)ᵀ`.
+    z = a \ transpose(x)
+    return x isa AbstractVector ? vec(z) : permutedims(z)
 end
 
 ### Algebra
--- a/src/pdsparsemat.jl
+++ b/src/pdsparsemat.jl
@@ -68,8 +68,17 @@ end
 *(a::PDSparseMat, c::Real) = PDSparseMat(a.mat * c)
 *(a::PDSparseMat, x::AbstractMatrix) = a.mat * x  # defining these seperately to avoid
 *(a::PDSparseMat, x::AbstractVector) = a.mat * x  # ambiguity errors
-\(a::PDSparseMat{T}, x::AbstractVecOrMat{T}) where {T <: Real} = convert(Array{T}, a.chol \ convert(Array{Float64}, x)) #to avoid limitations in sparse factorization library CHOLMOD, see e.g., julia issue #14076
-/(x::AbstractVecOrMat{T}, a::PDSparseMat{T}) where {T <: Real} = convert(Array{T}, convert(Array{Float64}, x) / a.chol)
+# `x` is converted to `Float64` to work around CHOLMOD limitations (julia issue #14076).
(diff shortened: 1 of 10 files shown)


src/pdsparsemat.jl
```diff
@@ -68,8 +68,17 @@ end
...
 *(a::PDSparseMat, x::AbstractVector) = a.mat * x  # ambiguity errors
-\(a::PDSparseMat{T}, x::AbstractVecOrMat{T}) where {T <: Real} = convert(Array{T}, a.chol \ convert(Array{Float64}, x)) #to avoid limitations in sparse factorization library CHOLMOD, see e.g., julia issue #14076
-/(x::AbstractVecOrMat{T}, a::PDSparseMat{T}) where {T <: Real} = convert(Array{T}, convert(Array{Float64}, x) / a.chol)
+# `x` is converted to `Float64` to work around CHOLMOD limitations (julia issue #14076).
+function \(a::PDSparseMat, x::AbstractVecOrMat{<:Real})
+    @check_argdims a.dim == size(x, 1)
+    T = promote_type(eltype(a), eltype(x))
+    return convert(Array{T}, a.chol \ convert(Array{Float64}, x))
+end
+function /(x::AbstractVecOrMat{<:Real}, a::PDSparseMat)
```
andreasnoack, line 77: Coverage suggests that this one is not hit

devmotion: AFAICT this method has never worked and is explicitly disabled in the tests.

devmotion: I added a workaround for the non-existing `/` in SparseArrays in https://github.com/JuliaStats/PDMats.jl/pull/236/commits/c814d220a6e93fa657515537fdcb3a99245b85c6

andreasnoack (approved):

PR #238, opened 2026-06-25
Decouple `PDMat`'s matrix and Cholesky factor types (#237)

### LINKED ISSUE #237: PDMat cannot pair a dense `mat` with a structured/lazy Cholesky factor (shared type parameter `S`)
## Summary

`PDMat{T,S}` constrains `mat` and the Cholesky factor to the **same** array type `S`:

```julia
struct PDMat{T <: Real, S <: AbstractMatrix{T}} <: AbstractPDMat{T}
    mat::S
    chol::Cholesky{T, S}
end
```

A `PDMat` therefore cannot be built from a dense `mat` paired with a `Cholesky` whose factor has a *different* array type — e.g. a lazy `Fill` factor produced by a structured/lazy `cholesky` overload.

## Reproducer (PDMats + FillArrays only)

```julia
using PDMats, LinearAlgebra, FillArrays

mat  = [2.0 0.0; 0.0 2.0]
chol = Cholesky(Fill(sqrt(2.0), 2, 2), 'U', 0)   # factor type ≠ typeof(mat)

PDMat(mat, chol)
# ERROR: Input array contains both 0.0 and 2.0. Cannot convert to Fill
```

`PDMat(mat, chol)` infers `S` from `chol.factors` (a `Fill`) and then stores the dense `mat` at that same type, which fails. Tested on PDMats 0.11.37.

## Real-world trigger

`SparseConnectivityTracer` (≥ 1.2.2) overloads `cholesky(::AbstractMatrix{<:AbstractTracer})` to return a conservative all-depends factor as a `Cholesky{T,<:Fill}`. Building a dense `MvNormal`/`PDMat` over a tracer covariance — e.g. when probing model sparsity in a downstream package — then hits the error above, because the dense covariance `mat` and the `Fill` factor have different types.

## Proposed fix

#188 already resolves this by decoupling the factorization type from `S`:

```julia
struct PDMat{T <: Real, S <: AbstractMatrix{T}, F <: Factorization} <: AbstractPDMat{T}
    mat::S
    fact::F
end
```

Since #188 is a larger change (it also unifies `PDMat`/`PDSparseMat` and generalizes to arbitrary `Factorization`s), would it be worth extracting just the type-decoupling part — a trailing type parameter for the factor, independent of `mat`'s type — into a smaller, **non-breaking** release?

The added parameter is trailing, so existing `PDMat{T,S}` and `PDMat{T,S} where {T,S}` signatures continue to match instances with the extra parameter left free, and the `PDMat{T,S}(mat, chol)` constructors / `.chol` access can be retained — i.e. the concrete type parameters are an implementation detail rather than public API.

devmotion:
> [!NOTE]
> **Builds on #239** (now merged). This PR was stacked on top of it and has been retargeted to `master`, so the diff shown here is only this PR's own changes.
>
> **Why #239 was necessary:** decoupling the matrix from the Cholesky factor type means `PDMat` now propagates the matrix's own type — e.g. a `Symmetric` wrapper — into `.mat`, instead of densifying it to a plain `Matrix`. Addition then forwards `.mat` straight into `_adddiag` (e.g. `+(a::PDMat, b::PDiagMat) = PDMat(_adddiag(a.mat, b.diag))`), which previously was only defined for `Union{Matrix, SparseMatrixCSC}` and would now hit a `MethodError`. #239 generalized `_adddiag`/`_scaleadddiag` to any `AbstractMatrix`, so this PR relies on it to keep `+`/`pdadd` working for the newly-preserved wrapper types.

Closes #237.

## Problem

`PDMat{T, S}` forced the stored matrix `mat::S` and its Cholesky factorization `chol::Cholesky{T, S}` to share one array type `S`. Constructing a `PDMat` from a dense matrix and a `Cholesky` whose factors have a *different* array type therefore failed, because the matrix was coerced to the factor type:

```julia
julia> PDMat(mat, chol)   # chol.factors isa Fill
ERROR: Input array contains both 0.0 and 2.0. Cannot convert to Fill
```

This broke dense `MvNormal`/`PDMat` construction over covariances produced by `SparseConnectivityTracer` (≥ 1.2.2), which returns `Cholesky{T, <:Fill}`.

## Change

Add a third type parameter `C <: Cholesky{T}` so the factorization's storage type is independent of the matrix's:

```julia
struct PDMat{T <: Real, S <: AbstractMatrix{T}, C <: Cholesky{T}} <: AbstractPDMat{T}
    mat::S
    chol::C
end
```

The two-argument constructor keeps both arguments as-is when their element types match (a fast path that preserves identity, `isbits`-ness, and allocation behavior) and promotes them otherwise — instead of the previous lossy "take the element type from `chol`" coercion.

`PDMat{T, S}` and `PDMat{T}` remain valid partial (`UnionAll`) types, so existing dispatch, construction, conversion, and the deprecations continue to work unchanged.

### Behavior change

Because the matrix is no longer coerced to the factor type, `PDMat` now keeps the matrix's own type. For example `PDMat(Symmetric(A))` keeps the `Symmetric` wrapper (the data is not densified), where previously `.mat` was the bare underlying matrix. The `specialarrays.jl` banded-matrix test is updated accordingly.

## Tests / misc

- Regression test pairing a dense matrix with a `Fill`-factored `Cholesky`, looping over `mat`/`chol` element-type combinations across `PDMat`, `PDMat{T}`, and `PDMat{T, S}`. `FillArrays` is added as a test dependency.
- Refreshed a stale doctest repr in `generics.jl`.

Full test suite passes locally.

### DIFF
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -37,6 +37,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           repository: ${{ matrix.package.user }}/${{ matrix.package.repo }}
+          ref: 'dmw/aliases'
           path: downstream
       - name: Load this and run the downstream tests
         shell: julia --color=yes --project=downstream {0}
--- a/.github/workflows/IntegrationTest.yml
+++ b/.github/workflows/IntegrationTest.yml
@@ -37,7 +37,6 @@ jobs:
         uses: actions/checkout@v6
         with:
           repository: ${{ matrix.package.user }}/${{ matrix.package.repo }}
-          ref: 'dmw/aliases'
           path: downstream
       - name: Load this and run the downstream tests
         shell: julia --color=yes --project=downstream {0}
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "PDMats"
 uuid = "90014a1f-27ba-587c-ab20-58faa44d9150"
-version = "0.11.38"
+version = "0.11.39"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "PDMats"
 uuid = "90014a1f-27ba-587c-ab20-58faa44d9150"
-version = "0.11.39"
+version = "0.11.40"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/Project.toml
+++ b/Project.toml
@@ -15,6 +15,7 @@ StatsBaseExt = "StatsBase"
 
 [compat]
 BandedMatrices = "0.15, 1"
+FillArrays = "1"
 FiniteDifferences = "0.12"
 ForwardDiff = "0.10, 1"
 Random = "<0.0.1, 1"
@@ -25,6 +26,7 @@ julia = "1.10"
 
 [extras]
 BandedMatrices = "aae01518-5342-5314-be14-df237901396f"
+FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
 FiniteDifferences = "26cc04aa-876d-5657-8c51-4c34ba976000"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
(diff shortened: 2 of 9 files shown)


devmotion:
I think this PR is ready for review now.

IMO the remaining 3 test failures in the Distributions tests are not practically relevant, they test whether some dense `MvNormal` distributions are subtypes of the `FullMvNormal` type aliases which relies on the type parameters of `PDMat` which arguably are not public API. This doesn't seem to affect anything in practice though apart from less compact printing: https://github.com/search?q=repo%3AJuliaStats%2FDistributions.jl+FullNormal&type=code

andreasnoack:
Wouldn't it be better to adjust the Distributions tests first to avoid breaking CI there?

devmotion:
Initially I thought it wouldn't be possible (unless we'd remove the tests) but I got an idea, let me check it.

devmotion:
CI shows that the Distributions downstream tests in this PR pass with https://github.com/JuliaStats/Distributions.jl/pull/2073 (for which Distributions tests with the latest PDMats release pass as well)

andreasnoack (approved):

PR #239, opened 2026-06-25
Support adding `AbstractPDMat`s backed by static arrays

### LINKED ISSUE #167: Missing Method for StaticArrays
Looks like there are still a few things hiding out that use `Matrix` instead of `AbstractMatrix`.

Here's one example:

```julia
julia> a=PDMat(SMatrix{3,3}([ 0.880022  0.646205  0.598234
        0.646205  0.578793  0.449772
        0.598234  0.449772  1.33728]))
3×3 PDMat{Float64, SMatrix{3, 3, Float64, 9}}:
 0.880022  0.646205  0.598234
 0.646205  0.578793  0.449772
 0.598234  0.449772  1.33728

julia> b = PDiagMat(SVector{3}([0.558713 0.384665 0.469067]))
3×3 PDiagMat{Float64, SVector{3, Float64}}:
 0.558713  0.0       0.0
 0.0       0.384665  0.0
 0.0       0.0       0.469067

julia> a+b
ERROR: MethodError: no method matching _adddiag(::SMatrix{3, 3, Float64, 9}, ::SVector{3, Float64})
```

However, `a+a` and `b+b` work. So does `a+b` when both are constructing from normal matrices.

devmotion:
Fixes #167.

`+` and `pdadd` of a `PDMat`/`PDiagMat`/`ScalMat` errored when the underlying storage was a static array, e.g.

```julia
PDMat(SMatrix{3,3}(...)) + PDiagMat(SVector{3}(...))
# ERROR: MethodError: no method matching _adddiag(::SMatrix{3,3,Float64,9}, ::SVector{3,Float64})
```

### Root cause
`_adddiag`/`_adddiag!` were only defined for `Union{Matrix, SparseMatrixCSC}` and mutated in place — which is impossible for immutable arrays (and `copy(::SMatrix)` is still immutable). Subtraction already worked because it routes through Base's broadcasting fallback.

### Changes
- Replace the non-mutating `_adddiag` wrappers with generic `AbstractMatrix` methods that build the result by construction (`a + Diagonal(v)`, `a + v*I`, `a .+ c.*Diagonal(v)`). Allocation-measured to match the previous copy-and-mutate path for dense matrices (1 allocation), stay sparse for sparse inputs, and preserve the static container otherwise.
- Add `_scaleadddiag` for the matrix-scaled `pdadd` paths (`c*a + diagonal`), keeping an in-place fast path for mutable storage (avoids a second allocation) and falling back to `muladd` for other matrix types.
- Keep the in-place `_adddiag!` — it is still used by the zero-allocation `pdadd!` and the dense `Matrix(b)` fallbacks.
- Vector-taking variants retain `@check_argdims` to guard against silent broadcast singleton-expansion.

### Tests
Added coverage in `test/specialarrays.jl` for every `+`/`pdadd` combination of `PDMat{SMatrix}` / `PDiagMat{SVector}` / `ScalMat`, asserting both numerical correctness and static-type preservation.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "PDMats"
 uuid = "90014a1f-27ba-587c-ab20-58faa44d9150"
-version = "0.11.38"
+version = "0.11.39"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/addition.jl
+++ b/src/addition.jl
@@ -55,18 +55,18 @@ if HAVE_CHOLMOD
     pdadd(a::PDMat, b::PDSparseMat, c::Real) = PDMat(a.mat + b.mat * c)
 end
 
-pdadd(a::PDiagMat, b::PDMat, c::Real) = PDMat(_adddiag!(b.mat * c, a.diag, one(c)))
+pdadd(a::PDiagMat, b::PDMat, c::Real) = PDMat(_scaleadddiag(b.mat, c, a.diag))
 pdadd(a::PDiagMat, b::PDiagMat, c::Real) = PDiagMat(a.diag + b.diag * c)
 pdadd(a::PDiagMat, b::ScalMat, c::Real) = PDiagMat(a.diag .+ b.value * c)
 if HAVE_CHOLMOD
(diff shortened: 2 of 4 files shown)


devmotion:
This is required for #238

andreasnoack (approved):

PR #245, opened 2026-09-21
cleanup SuiteSparse

araujoms:


devmotion:
The package just shouldn't depend on SparseArrays at all: #188

devmotion:
Can you explain what motivates this particular PR, what issue it fixes etc.?

araujoms:
Just reducing noise.

devmotion:
Reducing noise?

araujoms:
Yes. SuiteSparse doesn't exist anymore, so its presence in the codebase is just noise. It's nothing exciting but I think it's good to remove cobwebs now and then.

devmotion:
Sure, it's good to clean up stuff from time to time, I'm just confused why/how that reduces noise or what noise would be caused by the existing releases.

devmotion (approved):

araujoms:
Specifically, it caused `SuiteSparse` to appear as a dependency of `Hypatia`, through `Distributions`.

